PR #82, opened 2014-03-04
Array.isArray() and Object.keys() shims

ghost:
With these hacks, we can hopefully get to es3 compat and I'll switch over stream-browserify to use readable-stream directly instead, saving many bytes in mixed `require('stream')` vs `require('readable-stream')` usage.

### DIFF
--- a/build/files.js
+++ b/build/files.js
@@ -46,9 +46,21 @@ const requireReplacement = [
           /^(var util = require\('util'\);)/m
         , '$1\nvar isArray = require(\'isarray\');'
       ]
-    , isArray = [
-          /Array\.isArray/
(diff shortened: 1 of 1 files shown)


f0rmat1k:
You still need es3?

mcollina:
@f0rmat1k yes we do until we bump the major, as it is a semver-major change.

f0rmat1k:
@mcollina ok. Are you gonna to release such major? I can send PR. It's just rly strange to see in 2017 `isarray` dependency, when you install such pockets as mongodb for example.

mcollina:
@f0rmat1k at the moment there is no plan to release a semver major change any time soon.

Issue #124, opened 2015-04-03
Standardizing ".destroy()"


chrisdickinson:
This is a behavior that streams [_used_ to have](https://github.com/iojs/io.js/blob/v1.x/lib/stream.js#L68) (as of streams 1), but was never really fully specified, and was removed in [streams2+](https://github.com/iojs/io.js/blob/v1.x/lib/_stream_readable.js#L554-L563). The result is that there's no standard way to prematurely end a stream from the outside, and the effects of prematurely ended streams are undefined (or surprising to our users).

@mafintosh may have more words about this!

domenic:
FWIW: in WHATWG streams, we have `readableStream.cancel()` and `writableStream.abort()`. The former makes the readable stream look closed, whereas the latter makes the writable stream look errored. I can go into more detail on the reasoning if desired.

mafintosh:
I think an overridable `._destroy(cb)` would be good way to do this and then have all streams have a public `.destroy` that would call `._destroy` once and only once. When the callback is called the stream would emit `close` and possibly `error` if an error is passed.

``` js
myStream._destroy = function (cb) {
  closeResources(function (err) {
    cb(err)
  })
}

myStream.destroy() // calls the above method
myStream.destroy() // doesn't do anything
```

emilbayes:
+1! I always end up bike-shedding my own `destroy` method that for example cleans up the underlying source of `Readable`. However I've never really been sure in what state I leave the stream...

chrisdickinson:
@Domenic I'd definitely be interested in why prematurely ending a writable implies an error.

@mafintosh So, to make sure I've got this correct, a few of the potential event sequences are:

##### `A.pipe(B).pipe(C); A.destroy()` – streams2+, default _destroy 

1. User calls "A.destroy()".
2. "A" emits *"close"*.

A is still piped to B, which is still piped to C.

##### `A.pipe(B).pipe(C); B.destroy()` – streams2+, default _destroy

1. User calls "B.destroy()".
2. "B" emits *"close"*.
3. "A.unpipe(B)"
  1. "B" emits *"unpipe" A*.
  2. If "B" is the only destination of "A", "A" is implicitly paused.
  3. Else if "B" awaitsDrain, then "A" may start flowing.

B is still piped to C. A is still available to be re-piped, and may be paused or flowing.

##### `A.pipe(B).pipe(C); B.destroy()` – A is streams1, B and C are streams2+, default _destroy

1. User calls "B.destroy()".
[...]

PR #125, opened 2015-04-09
Add .destroy to Readable and Writable prototypes

mafintosh:
This is an attempt to implement #124. Basically it adds `.destroy` to the Readable and Writable prototypes. These prototypes forwards the call to `._destroy(cb)` the first time destroy is called. If destroy is called twice the second call is ignored.

This allows you prematurely end a stream more easily. As an example here is a file reader

``` js
var createFileStream = function (fd) {
  var rs = new stream.Readable()

  rs._destroy = function (cb) {
    fs.close(fd, cb)
  }

  rs._read = function (size) {
    var buf = new Buffer(size)
    fs.read(fd, buf, 0, size, null, function (err, len) {
      if (err) {
        rs.emit('error', err)
        return rs.destroy()
      }
      rs.push(buf.slice(0, len))
    })
  }

  return rs
}
``` 

Calling `rs.destroy()` will now close the file descriptor and emit `close` when the close is done.

### DIFF
--- a/lib/_stream_readable.js
+++ b/lib/_stream_readable.js
@@ -108,6 +108,8 @@ function ReadableState(options, stream) {
     this.decoder = new StringDecoder(options.encoding);
     this.encoding = options.encoding;
   }
+
+  this.destroyed = false;
 }
 
 function Readable(options) {
@@ -150,6 +152,28 @@ Readable.prototype.unshift = function(chunk) {
 
 Readable.prototype.isPaused = function() {
   return this._readableState.flowing === false;
+};
+
+Readable.prototype.destroy = function() {
+  var self = this;
+  if (this._readableState.destroyed) return;
+  this._readableState.destroyed = true;
+
+  // if this is a duplex stream mark the writable part as destroyed as well
+  if (this._writableState) {
+    this._writableState.destroyed = true;
+  }
+
+  this._destroy(function(err) {
+    if (err) {
+      self.emit('error', err);
+    }
+    self.emit('close');
+  });
+};
+
+Readable.prototype._destroy = function (callback) {
+  process.nextTick(callback)
 };
 
 function readableAddChunk(stream, state, chunk, encoding, addToFront) {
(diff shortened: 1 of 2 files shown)


lib/_stream_readable.js
```diff
@@ -152,6 +154,28 @@ Readable.prototype.isPaused = function() {
...
 
+Readable.prototype.destroy = function() {
+  var self = this;
+  if (this._readableState.destroyed) return;
+  this._readableState.destroyed = true;
+
+  // if this is a duplex stream mark the writable part as destroyed as well
+  if (this._writableState) {
+    this._writableState.destroyed = true;
+  }
```
chrisdickinson: Sort of tempted to say "may as well specially implement `destroy` on Duplex" just to avoid adding another case of the parent class having intimate knowledge of its subclass.

chrisdickinson:
Curious: is the omission of `.destroy` from `Readable#pipe/on{error,close}` intentional?

lib/_stream_readable.js
mafintosh: yea i think thats good idea. will update the pr

mafintosh:
@chrisdickinson I forgot about that :) I'll add it.

lib/_stream_readable.js
sonewman: It's almost a shame we couldn't add this to the base `stream` class somehow since they are both doing pretty much the same thing.

Better still if there was a base `StreamState` that all states inherited from, although that would restrict people from calling `_destroy()` directly, although that might not be such a bad thing...

sonewman:
What do you think about adding it to the construction also? e.g.:
```javascript
var readable = new stream.Readable({
  destroy() {
    // burn down the house
  }
});
```

calvinmetcalf:
Shouldn't this be against the main io repo or is this just for discussion?

yoshuawuyts:
What steps must be taken to get this into core?

calvinmetcalf:
open pull request on this repo https://github.com/nodejs/node

calvinmetcalf:
I'm going to close this, an issue is where we probably want to continue this discussion.

PR #130, opened 2015-05-07
update with work done on io.js 2.0.0

slang800:
includes:
- most removal of is* functions
- renaming requires for '../common.js' to '../common'
- add simpler stream constructon (without subclassing) + tests
- remove a duplicated expression in lib/_stream_readable
- remove unused variables in lib/_stream_writable
- use nop as write() callback if omitted in lib/_stream_writable

I omitted the changes requiring process.nextTick to accept multiple args, and the addition of `_stream_wrap` for now.

### DIFF
--- a/lib/_stream_readable.js
+++ b/lib/_stream_readable.js
@@ -121,6 +121,9 @@ function Readable(options) {
   // legacy
   this.readable = true;
 
+  if (options && typeof options.read === 'function')
+    this._read = options.read;
+
   Stream.call(this);
 }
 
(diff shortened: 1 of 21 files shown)


calvinmetcalf:
see #121, I need to verify that works with 2.0.0 but it should in theory

calvinmetcalf:
(didn't mean to close)

calvinmetcalf:
you can use [io-stream](https://github.com/calvinmetcalf/io-stream) in the mean while to get the more recent goodies

calvinmetcalf:
see #135 which includes up to 2.2.1

calvinmetcalf:
done

PR #135, opened 2015-06-08
Merge io stream back in

calvinmetcalf:
merges the changes I made when publishing it as [io-stream](https://github.com/calvinmetcalf/io-stream) back in, contains the version of streams from node 2.2.1 tested to be working in node going back to v.0.8.

Outstanding issues include versioning, should this package be versioned to match the version of node/iojs it comes from or should it be versioned to independently?


dcousens:
I'll help out reviewing this,  will take a little bit though.

build/files.js
```diff
@@ -98,10 +98,47 @@ const headRegexp = /(^module.exports = \w+;?)/m
         + 'if (!EE.listenerCount) EE.listenerCount = function(emitter, type) {\n'
         + '  return emitter.listeners(type).length;\n'
         + '};\n/*</replacement>*/\n'
+      ]
```
calvinmetcalf: @dcousens the changes to this file is the meat of the update, most of the other changes are consequences of this

examples/typer-fsr.js
```diff
@@ -1,15 +0,0 @@
-var fs = require('fs');
```
calvinmetcalf: this is related to the fs file

float.patch
```diff
@@ -1,923 +0,0 @@
-diff --git a/lib/_stream_duplex.js b/lib/_stream_duplex.js
```
calvinmetcalf: it is very unclear why this was in the repo

fs.js
```diff
@@ -1,1705 +0,0 @@
-// Copyright Joyent, Inc. and other Node contributors.
```
calvinmetcalf: from what I can tell this is an (extremely) out of date version of the node fs library rewritten to use readable streams, unclear if it would even work as it hasn't been updated in a long time

readable.js
```diff
@@ -1,6 +1,9 @@
-var Stream = require('stream'); // hack to fix a circular dependency issue when used with browserify
+(function (){
+  try {
+    exports.Stream = require('st' + 'ream'); // hack to fix a circular dependency issue when used with browserify
```
calvinmetcalf: this was explicitly added to to avoid unnecessarily bundling the stream module

zlib.js
```diff
@@ -1,452 +0,0 @@
-// Copyright Joyent, Inc. and other Node contributors.
```
calvinmetcalf: same as the fs file this is an old version of zlib rewritten to include readable streams, looks like it was maybe part of a half finished effort to include all the places in the stdlib that included streams

build/files.js
```diff
@@ -98,10 +98,47 @@ const headRegexp = /(^module.exports = \w+;?)/m
...
+
+    , constReplacement = [
+        /const/g
+      , 'var'
```
calvinmetcalf: this replaces all the uses of const with var for compatibility reasons

build/files.js
```diff
@@ -98,10 +98,47 @@ const headRegexp = /(^module.exports = \w+;?)/m
...
+      ]
+
+    , bufferIsEncodingReplacement = [
+      /Buffer.isEncoding\((\w+)\)/
```
calvinmetcalf: replaces a single instance of Buffer.isEncoding with a manual check due to older versions of node not having the method

build/files.js
```diff
@@ -98,10 +98,47 @@ const headRegexp = /(^module.exports = \w+;?)/m
...
+    ,   '([\'hex\', \'utf8\', \'utf-8\', \'ascii\', \'binary\', \'base64\',\n'
+      + '\'ucs2\', \'ucs-2\',\'utf16le\', \'utf-16le\', \'raw\']\n'
+      + '.indexOf(($1 + \'\').toLowerCase()) > -1)'
+    ]
+
+    , requireStreamReplacement = [
+      /var Stream = require\('stream'\);/
```
calvinmetcalf: great pains are made to make sure `readableStream instanceof nativeStream` is true, this avoids bundling the streams module unless it's already included.  

In other words since you can't  `readableStream instanceof nativeStream` without first doing `var nativeStream = require('stream');` we only need to inherit from native streams if it's already bundled.

build/files.js
```diff
@@ -98,10 +98,47 @@ const headRegexp = /(^module.exports = \w+;?)/m
...
+    ,  '\n\n/*<replacement>*/\n'
+      + 'var Stream;\n (function (){try{\n'
+      + 'Stream = require(\'st\' + \'ream\');\n'
+      + '}catch(_){Stream = require(\'events\').EventEmitter;}}())'
+      + '\n/*</replacement>*/\n'
+    ]
+
+    , isBufferReplacement = [
+      /(\w+) instanceof Buffer/
```
calvinmetcalf: buffer instanceof Buffer doesn't work in the browser, so replaced with Buffer.isBuffer

build/files.js
```diff
@@ -98,10 +98,47 @@ const headRegexp = /(^module.exports = \w+;?)/m
...
+    , 'Buffer.isBuffer($1)'
+    ]
+
+    , processNextTickImport = [
+      headRegexp
+    , '$1\n\n/*<replacement>*/\nvar processNextTick = require(\'process-nextick-args\');\n/*</replacement>*/\n'
```
calvinmetcalf: these next 2 changes deal with process.nextTick which was updated to accept arguments

mafintosh:
Lots of changes :) This is overall looking fine to me. :+1: @calvinmetcalf

calvinmetcalf:
yes there are lots, though everything in the lib and test directory is auto generated

calvinmetcalf:
based on #136 I'm leaning towards publishing this as v2.0.0 sometime tomorrow unless I hear objections

calvinmetcalf:
updated with some changes to the readme to explain versioning and list the wg members, (do speak up if I missed you)

dcousens:
Will review in the next 10 hours.  If I don't ACK/comment by then,  assume good to go and I'll post review.

calvinmetcalf:
ok forced push to get the tests working

calvinmetcalf:
@dcousens sorry I could only gave you 8 hours

dcousens:
No worries @calvinmetcalf, I'll review post-haste.   Sorry.

dcousens:
LGTM

PR #139, opened 2015-06-10
browserifiable tests

calvinmetcalf:
best idea I have so far is hand porting tests over to tape


package.json
```diff
@@ -5,17 +5,19 @@
...
-    "process-nextick-args": "~1.0.0",
     "inherits": "~2.0.1",
     "isarray": "0.0.1",
+    "process-nextick-args": "~1.0.0",
     "string_decoder": "~0.10.x",
     "util-deprecate": "~1.0.1"
   },
   "devDependencies": {
-    "tap": "~0.2.6"
+    "tap": "~0.2.6",
```
mafintosh: why is tap needed?

calvinmetcalf: tap is what is used to run the current tests from node, it's somewhat idiosyncratic how it runs at it looks like it just emits tap output via console logs so thats how it's interpreted, I suppose we could just run all the scrypts and feed it to tspec or something

calvinmetcalf:
ok all of them are ported over and when running in the browser manually all of them pass except 

- test-stream-writable-decoded-encoding
- test-stream-writev

and they all fail due to string decoding issues

calvinmetcalf:
found the issue I forgot to add the g flat to the regex which replaced `foo instanceof Buffer` with `Buffer.isBuffer`, meaning this pull is likely going to end up being 2.0.1

calvinmetcalf:
and added sauce labs

calvinmetcalf:
most of the saucelabs config taken directly from what @feross set up for buffer

package.json
mafintosh: you can run the tests with tape as well `tape test/parallel/*.js`

calvinmetcalf:
ugh, the problem is that I accidentally picked free instead of open source when making the saucelabs account and the site seems to be having a bug preventing you from change, emailing their tech support now

calvinmetcalf:
ok fixed by saucelab running the tests again

package.json
calvinmetcalf: as it turns out, no you can't, not in 0.8

calvinmetcalf:
Ugh opera 11 is probably unnecessary

calvinmetcalf:
for some reason the iphone/ipad tests are just hanging pretty early on removing them and everything passes

calvinmetcalf:
It works! now to rebase
yata

calvinmetcalf:
ah I was testing *really* old iThings

calvinmetcalf:
rebased, ready for @nodejs/streams to take a look

build/common-replacements.js
```diff
@@ -26,3 +26,16 @@ module.exports.altIndexOfUseReplacement = [
     /(\W)([\w\.\(\),\[\]]+)(\.indexOf\()/gm
   , '$1indexOf($2, '
 ]
+module.exports.objectKeysDefine = [
```
calvinmetcalf: moved this out into the common files so that it could also be used on tests

build/files.js
```diff
@@ -63,10 +63,15 @@ const headRegexp = /(^module.exports = \w+;?)/m
       ]
 
     , objectDefinePropertyReplacement = [
-          /Object.defineProperties/
+          /(Object\.defineProperties)/
```
calvinmetcalf: I'm not 100% sure this is used anywhere as the previous version didn't actually work

build/files.js
```diff
@@ -63,10 +63,15 @@ const headRegexp = /(^module.exports = \w+;?)/m
...
       ,   'if (Object.defineProperties) $1'
       ]
-
+    , objectDefinePropertySingReplacement = [
```
calvinmetcalf: IE 8 has an Object.defineProperty function, it just doesn't work

build/files.js
```diff
@@ -122,7 +117,7 @@ const headRegexp = /(^module.exports = \w+;?)/m
     ]
 
     , isBufferReplacement = [
-      /(\w+) instanceof Buffer/
+      /(\w+) instanceof Buffer/g
```
calvinmetcalf: the lack of `g` here caused a bunch of bugs with encoding and is why we probably want to publish 2.0.1 soon.

build/test-replacements.js
```diff
@@ -106,6 +112,18 @@ module.exports['common.js'] = [
         + '}\n'
         + '/*</replacement>*/\n'
     ]
+  , [
```
calvinmetcalf: this isn't needed in the browser and cause some issues with browsers that don't have uint8clampedarray

build/test-replacements.js
```diff
@@ -106,6 +112,18 @@ module.exports['common.js'] = [
...
+        /^if \(global\.ArrayBuffer\) \{([^\}]+)\}$/m
+      ,   '/*<replacement>*/if (!process.browser) {'
+        + '\nif \(global\.ArrayBuffer\) {$1}\n'
+        + '}/*</replacement>*/\n'
+    ]
+    , [
+          /^Object\.defineProperty\(([\w\W]+?)\}\)\;/mg
```
calvinmetcalf: same here, this defines some stuff we never actually use and causes issues in older browsers that don't have a working copy of Object.defineProperty

lib/_stream_duplex.js
```diff
@@ -21,6 +14,13 @@ var objectKeys = Object.keys || function (obj) {
 /*</replacement>*/
 
 
+module.exports = Duplex;
```
calvinmetcalf: since the tests don't have an export I switched the objectKeys replacement to hook onto the `'use strict';` statement which is now in every file.

test/browser.js
```diff
@@ -0,0 +1,61 @@
+if (!global.console) {
```
calvinmetcalf: IE9 only passed if I had the Dev tools open without this, because console isn't an actual object unless you have them open.

test/browser/test-stream2-compatibility.js
```diff
@@ -0,0 +1,33 @@
...
+
+    inherits(TestReader, R);
+
+    TestReader.prototype._read = function(n) {
+      this.push(this._buffer);
+      this._buffer = new Buffer(0);
+    };
+
+    var reader = new TestReader();
+    setTimeout(function() {
```
calvinmetcalf: this was originally setImmediate but IE has a broken version of setImmediate in all versions which prevented tests from passing.

calvinmetcalf:
one last change, I re-enabled opera 11

calvinmetcalf:
the saucelabs task was right up against the limit of the maximum lenght of a build so sometimes timed out this tries breaking it into multiple tasks

calvinmetcalf:
splinting them up seems to work

calvinmetcalf:
[Sauce Test Status](https://saucelabs.com/u/readable-stream)

dcousens:
@calvinmetcalf this is massive.  Well done.

test/common.js
```diff
@@ -11,6 +11,15 @@ if (!global.clearImmediate) {
 }
 /*</replacement>*/
 'use strict';
+
+/*<replacement>*/
```
dcousens: Why not just `Object.keys`?

build/common-replacements.js
dcousens: Why is it necessary at all?

build/files.js
dcousens: @calvinmetcalf why are all these regex replacements necessary anyway?

build/common-replacements.js
calvinmetcalf: Because some browsers don't have Object.keys and we want to test in them

build/files.js
calvinmetcalf: Yeah so this one replaces all use of foo instanceof buffer with Buffer.isBuffer(foo)  because the former will not work in the browser but the latter will.  This caused issues with the encoding argument for write and transform

calvinmetcalf:
@dcousens the regexs are to replace node specific study with things that work in the brieee

calvinmetcalf:
Didn't mean to close,  things that work across browsers

build/common-replacements.js
dcousens: Any of [these browsers](https://cloud.githubusercontent.com/assets/1128607/8131283/3a814a02-10e7-11e5-85c0-8d3887303f70.png)?

As far as I could tell on https://kangax.github.io/compat-table/es5/,  all the above browsers supported `Object.keys`, only if we go as far back as `IE8`, `FF 3.5` or `Safari 4`, does that actually become an issue.

feross: FWIW, in theory buffer works back to IE6. Tests used to pass but they
became flakey a while ago and I thought it was SauceLab's fault so I just
disabled IE6 and IE7 instead of debugging it. Tests still pass down to IE8.

The browserify http module claims to work down to IE5.5!

calvinmetcalf: It showed up in a really old version of mobile safari, I ended up not going
back as far but figured avoiding the Object.keys thing in that test file
could hurt

dcousens:
@calvinmetcalf can we just plug an [ES5 shim](https://www.npmjs.com/package/es5-shim) in for the tests rather than doing it ourselves?
You've already shown in the errors it contains that it is not a simple task by any means.

calvinmetcalf:
so the object keys one was already there being done internally, overall we could use the shim if we wanted to test IE8 support but otherwise I haven't added all that much

dcousens:
> so the object keys one was already there being done internally

Sure, but we can do better IMHO.

dcousens:
> if we wanted to test IE8 support

Do we?  I mean,  the browsers listed all seem to support ES5 / ES5.1 no problem, so I'm not sure why we need these shims at all?

calvinmetcalf:
rebased and rebuilt the docs for version 2.3.0

calvinmetcalf:
@dcousens I didn't add any shims I just moved one we were already using to the common files so we could do it for the test file as well, the main issue was that the testcommons.js file was throwing an error when Object.keys wasn't available that the library files didn't because that was regexed out. Otherwise the tests pass without any sort of extra shim added, I debated adding support for testing IE8 but that required a bunch of extra stuff added (Array.prototype.some for something in the tests and then Function.prototype.bind for the actual library) so that seemed a bit much).

calvinmetcalf:
I'd love to get this merged soon, it's the blocker for releasing 2.0.1 with the fix for the incorrect encoding and not exporting .Stream anyone able to give a :+1:

build/files.js
dcousens: Why not just permanently replace `instanceof` with `Buffer`?

dcousens:
> I debated adding support for testing IE8 but that required a bunch of extra stuff added (Array.prototype.some for something in the tests and then Function.prototype.bind for the actual library) so that seemed a bit much).

It seems odd you have shimmed some things, but not others.  Why not just use modern steps and let the shim be dropped in if support isn't as expected?

calvinmetcalf:
because Object.keys was already replaced by an objectKeys function in the library code, all I did was make sure it also was being done so in the tests, ` Function.prototype.bind ` is used extensively in the library so I didn't want to shim that one **right now** because I'd probably have to do it globally.

I'm not against IE8 support I just wanted to punt it into a latter issue.

build/files.js
calvinmetcalf: because this is code that is coming in from io.js where they use instanceof so something to take up with them

sonewman:
this seems like it would be good to merge, any other thoughts @nodejs/streams ?

calvinmetcalf:
going to merge this, i think @dcousens's issues are mainly questions about browser support which we can address in the future easily

calvinmetcalf:
b6810590821e2082d5b79e3e0d08d3d090ebc353

dcousens:
> going to merge this, i think @dcousens's issues are mainly questions about browser support which we can address in the future easily

Correct, wasn't meaning to block the merge.

PR #142, opened 2015-06-13
add docs

calvinmetcalf:
inspired by #100


dcousens:
LGTM.

However, are these docs going to be maintained here from now on?  Are the docs going to be synchronized?

calvinmetcalf:
Yes I added it to the build process

Issue #143, opened 2015-06-14
Stream export missing in 2.0.0


dougwilson:
Hi! It looks like the commit d0d888fc86cb30d33ca45f7adad3ac68e7a9afbe may have accidentally broken the `Stream` export. This is because it's assigning `exports.Stream` first, but then re-assigning the value of `exports` right after, so `exports.Stream` is now missing.

Example
```js
require('readable-stream').Stream // undefined after the above the commit
```

calvinmetcalf:
Woops we probably want to set exports.Stream to be the same as module.exports if we can't import stream

dougwilson:
In this case, importing `stream` works fine, bu the order of the lines are stomping on each other :)

Instead of
```js
(function (){
  try {
    exports.Stream = require('st' + 'ream'); // hack to fix a circular dependency issue when used with browserify
  } catch(_){}
}());
exports = module.exports = require('./lib/_stream_readable.js');
```

they need to be in the other order:

```js
exports = module.exports = require('./lib/_stream_readable.js');
(function (){
  try {
    exports.Stream = require('st' + 'ream'); // hack to fix a circular dependency issue when used with browserify
  } catch(_){}
}());
```

calvinmetcalf:
Yes the other thing is for use in browserify.

dougwilson:
Thank you, 2.0.1 fixes the issue :)

PR #144, opened 2015-06-15
correctly export .Stream

calvinmetcalf:
fix for #143

### DIFF
--- a/readable.js
+++ b/readable.js
@@ -4,7 +4,7 @@ var Stream = (function (){
   } catch(_){}
 }());
 exports = module.exports = require('./lib/_stream_readable.js');
-exports.Stream = Stream || module;
+exports.Stream = Stream || exports;
 exports.Readable = exports;
 exports.Writable = require('./lib/_stream_writable.js');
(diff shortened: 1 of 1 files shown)


readable.js
```diff
@@ -1,9 +1,10 @@
-(function (){
+var Stream = (function (){
   try {
-    exports.Stream = require('st' + 'ream'); // hack to fix a circular dependency issue when used with browserify
+    return require('st' + 'ream'); // hack to fix a circular dependency issue when used with browserify
   } catch(_){}
 }());
 exports = module.exports = require('./lib/_stream_readable.js');
+exports.Stream = Stream || exports;
```
dcousens: `exports.Stream = exports`

So, you could have `exports.Stream.Stream.Stream.Stream === exports`, wait what?

calvinmetcalf: it's exactly the same situation you have currently with exports.Readable on the next line

dcousens: All good,  just struck me as odd at first, but you're indeed right.  This is exactly mimicking the node export.

dcousens:
ACK LGTM.

PR #146, opened 2015-06-19
use container-based infrastructure for faster build

shinnn:
http://docs.travis-ci.com/user/workers/container-based-infrastructure/

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,3 +1,4 @@
+sudo: false
 language: node_js
 before_install:
   - npm install -g npm


calvinmetcalf:
do you mind rebasing it off of #139 to make sure it works there ?

dcousens:
@shinnn it seems odd this isn't the default behaviour...

calvinmetcalf:
I think the idea is that it will be at some point

calvinmetcalf:
e3bf6c54e98b2241020be0d84721e5254214b84e

shinnn:
:+1:

shinnn:
@dcousens 

> it seems odd this isn't the default behaviour...

`sudo: false` by default breaks a lot of builds since various projects include `sudo apt get [something]` in their build scripts.

dcousens:
@shinnn then maybe the script should look for `sudo` and assume `sudo: false` if it doesn't exist?

shinnn:
Good idea! :point_right: https://github.com/travis-ci/travis-ci/issues

PR #150, opened 2015-07-02
browserify compatibility

zertosh:
Fixes https://github.com/substack/node-browserify/issues/1312. After this I can bump stream-browserify to ^2.0.0 in browserify.

### DIFF
--- a/lib/_stream_readable.js
+++ b/lib/_stream_readable.js
@@ -32,7 +32,7 @@ if (!EE.listenerCount) EE.listenerCount = function(emitter, type) {
 var Stream;
  (function (){try{
 Stream = require('st' + 'ream');
-}catch(_){Stream = require('events').EventEmitter;}}())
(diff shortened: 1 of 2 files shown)


zertosh:
The travis failures seem to be a saucelabs credentials issue

calvinmetcalf:
I'm away from a computer for the next two weeks so can't merge this or publish but it looks good so of @rvagg or @isaacs could merge and publish it that would be very helpful

dcousens:
LGTM

calvinmetcalf:
so the regex actually has to be updated not the result of it so I opened #153 which should be doing exactly the same this as this just via the updated regex

calvinmetcalf:
merged and published as 2.0.2

zertosh:
@calvinmetcalf Ah ok. Thanks!

PR #152, opened 2015-07-16
browserify compatibility

calvinmetcalf:
exactly like #150 but with an internal pull so hopefully the tests work


calvinmetcalf:
closing in favor of #153

PR #154, opened 2015-07-16
WIP Iojs 2.3.4

calvinmetcalf:
work in progress for iojs 2.3.4, nothing functional has changed just 'internal/util' is now used for depreciate, also a test for stream wrap was added, we will need to update some regexes


calvinmetcalf:
superseeded by #156

PR #156, opened 2015-08-03
upgrade to iojs 4.0.0

calvinmetcalf:

PR #158, opened 2015-09-18
Move require out of constructor to top of module

borisovg:
This provides a big performance improvement.

The Stream module code in current Node no longer has this issue.

### DIFF
--- a/lib/_stream_writable.js
+++ b/lib/_stream_writable.js
@@ -37,6 +37,8 @@ var Stream;
 
 var Buffer = require('buffer').Buffer;
 
+var Duplex = require('./_stream_duplex');
+
 util.inherits(Writable, Stream);
 
 function nop() {}
@@ -49,8 +51,6 @@ function WriteReq(chunk, encoding, cb) {
 }
 
 function WritableState(options, stream) {
-  var Duplex = require('./_stream_duplex');
-
   options = options || {};
 
   // object stream flag to indicate whether or not this stream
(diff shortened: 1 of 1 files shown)


borisovg:
## Benchmark results:

```
100000 reps in 12735524135 ns (7852 ops/s)

vs

100000 reps in 3121319649 ns (32037 ops/s)
```

## Test environment:

```
{ http_parser: '2.5.0',
  node: '4.0.0',
  v8: '4.5.103.30',
  uv: '1.7.3',
  zlib: '1.2.8',
  ares: '1.10.1-DEV',
  modules: '46',
  openssl: '1.0.2d' }
```

## Benchmark code

```
var Writable = require('readable-stream/writable');

var i, startTime,
    reps = 100000,
    semaphore = reps;

function _make_socket () {
    var data = [],
        ws = new Writable();

    ws._write = function (chunk, encoding, next) {
        data.push(chunk.toString());
        next();
    };

    ws.on('finish', function () {
        var d;

        d = data.join('');

        semaphore -= 1;

        if (!semaphore) {
            end(startTime);
        }
    });

    return ws;
}

startTime = process.hrtime();

for (i = 0; i < reps; i += 1) {
    var ws = _make_socket();

    ws.end(new Buffer('foo'));
}

function end (startTime) {
    var diff = process.hrtime(startTime);
    var ns = diff[0] * 1e9 + diff[1];
    var s = diff[0] + diff[1] * 1e-9;

    console.log('%d reps in %d ns (%d ops/s)', reps, ns, Math.floor(reps / s));
    console.log(process.versions);
}
```

calvinmetcalf:
So this is actually a set of regexs to convert the node stuff to work in the browser, let me look into how we could do this on top of #156

calvinmetcalf:
wait nope there is actually a very good reason for that to be done that way, naimly to avoid a circular dependency

calvinmetcalf:
so this breaks transform streams and likely enables that speedup by breaking something, that being said we might be able to cache the require call, though this might be an issue with browserify not doing so itself

PR #160, opened 2015-09-19
fix issue with upgrading npm directly to 3.0.0 on node 0.8

calvinmetcalf:
fix for #158, built off of #156

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,6 +1,7 @@
 sudo: false
 language: node_js
 before_install:
+  - npm install -g npm@2
   - npm install -g npm
 notifications:
   email: false
--- a/test/common.js
+++ b/test/common.js
@@ -32,6 +32,7 @@ exports.libDir = path.join(exports.testDir, '../lib');
 exports.tmpDirName = 'tmp';
 exports.PORT = +process.env.NODE_COMMON_PORT || 12346;
 exports.isWindows = process.platform === 'win32';
+exports.isAix = process.platform === 'aix';
 
 function rimrafSync(p) {
   try {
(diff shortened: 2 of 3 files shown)


calvinmetcalf:
but wait .travis is now suddenly failing for some unrelated reason

calvinmetcalf:
part of this seems to be an npm issue npm/npm#9668 not sure about sauce labs

calvinmetcalf:
and that require has a very good reason for being inside the constructor, it's because it avoids a circular dependency

calvinmetcalf:
ping @nodejs/streams @dcousens

mafintosh:
@calvinmetcalf where is that `require` line you're mentioning above? Otherwise LTGM

.travis.yml
```diff
@@ -1,6 +1,7 @@
 sudo: false
 language: node_js
 before_install:
+  - npm install -g npm@2
```
dcousens: What are the implications of having both these lines?

test/common.js
```diff
@@ -32,6 +32,7 @@ exports.libDir = path.join(exports.testDir, '../lib');
 exports.tmpDirName = 'tmp';
 exports.PORT = +process.env.NODE_COMMON_PORT || 12346;
 exports.isWindows = process.platform === 'win32';
+exports.isAix = process.platform === 'aix';
```
dcousens: What is this used by?

calvinmetcalf: no idea, this is from node.js and was added between 4.0.0 and 4.1.0

calvinmetcalf:
@mafintosh that was here https://github.com/nodejs/readable-stream/blob/c2d1cce2a5000507c18481f703d28fe51ad7aab2/lib/_stream_writable.js#L59

.travis.yml
calvinmetcalf: so the version of npm that comes with node 0.8 breaks when it upgrades directly to npm@3, so it has to first upgrade to npm@2 and then from npm@2 upgrade to npm@3

dcousens:
:+1:

PR #161, opened 2015-09-22
Do not overwrite module var

nsimmons:
Minor change here to prevent a variable pointing to a module from being overwritten.

https://github.com/facebook/fbjs/blob/master/scripts/babel/inline-requires.js#L74-L91 checks for this and throws an error. 

`fbjs-scripts` module is used by https://github.com/facebook/react-native and this is causing tests to fail. 

This is one angle I am attempting to fix this issue. Why `fbjs-scripts` is throwing an error in this case is another matter that I will look into as well. This case may very well exist in other node modules. But in this specific case it was an easy change with basically no impact. 

Please consider accepting this change. Thanks!

### DIFF
--- a/build/files.js
+++ b/build/files.js
@@ -49,9 +49,10 @@ const headRegexp = /(^module.exports = \w+;?)/m
 
     , debugLogReplacement = [
           /var debug = util.debuglog\('stream'\);/
-      ,   '\n\n/*<replacement>*/\nvar debug = require(\'util\');\n'
-        + 'if (debug && debug.debuglog) {\n'
-        + '  debug = debug.debuglog(\'stream\');\n'
+      ,   '\n\n/*<replacement>*/\nvar debugUtil = require(\'util\');\n'
+        + 'var debug;\n'
+        + 'if (debugUtil && debugUtil.debuglog) {\n'
+        + '  debug = debugUtil.debuglog(\'stream\');\n'
         + '} else {\n'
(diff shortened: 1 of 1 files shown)


calvinmetcalf:
this needs to be fixed at [this location](https://github.com/nodejs/readable-stream/blob/c2d1cce2a5000507c18481f703d28fe51ad7aab2/build/files.js#L50-L58) due to how the module is setup.

More generally I'm not sure this will actually fix the base issue which is that fbjs is making some very unsafe assumptions when it comes to code, namely that requires will never be renamed in any way which is likely not safe for them to make.

cpojer:
See https://github.com/facebook/fbjs/issues/58

This is a bad jest setup on react-native and I'll fix it there. Nevertheless, require's *should* be using `const` (which is what import/export does in ES2015) so this fix is probably still appropriate :)

calvinmetcalf:
that's not something you can really count on in CJS contexts, (the underlying stream package DOES use const but it replaced by vars here for cross browser usage).

cpojer:
Sure, that's why I said *should*. It will make it easier to eventually move to ES2015 modules.

nsimmons:
Updated to make the fix in the right place.

calvinmetcalf:
@nsimmons thanks!

calvinmetcalf:
closing in favor of #163

PR #163, opened 2015-10-12
Nsimmons do not overwrite module var

calvinmetcalf:


calvinmetcalf:
rebased version of #161

PR #164, opened 2015-10-12
build for 4.1.2

calvinmetcalf:


calvinmetcalf:
and we are green @nodejs/streams would love to merge this and publish this as a minor version soon

lib/_stream_readable.js
```diff
@@ -18,7 +18,7 @@ var Buffer = require('buffer').Buffer;
 
 Readable.ReadableState = ReadableState;
 
-var EE = require('events').EventEmitter;
+var EE = require('events');
```
yoshuawuyts: won't this break on older versions of Node since we're relying on Node's provided module? Or is this comment moot since this is the result of the regex pull from upstream?

calvinmetcalf: we do [test it](https://travis-ci.org/nodejs/readable-stream/builds/84955012) against 0.8.x

lib/_stream_readable.js
```diff
@@ -18,7 +18,7 @@ var Buffer = require('buffer').Buffer;
...
 
 /*<replacement>*/
 var EElistenerCount = function(emitter, type) {
```
calvinmetcalf: @yoshuawuyts we never actually use it due to this line

mafintosh:
:+1:

PR #170, opened 2015-11-19
update to node 5.1.0

calvinmetcalf:
only real updates are to the docs

### DIFF
--- a/doc/stream.markdown
+++ b/doc/stream.markdown
@@ -3,8 +3,7 @@
     Stability: 2 - Stable
 
 A stream is an abstract interface implemented by various objects in
-Node.js.  For example a [request to an HTTP
-server](https://nodejs.org/docs/v5.0.0/api/http.html#http_http_incomingmessage) is a stream, as is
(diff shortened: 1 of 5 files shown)


calvinmetcalf:
rebased this to be based on #171 so we can get the perf boost that that pull brings, ready for review @nodejs/streams

calvinmetcalf:
add a commit to include the result of calvinmetcalf/process-nextick-args#5 which should speed things up

calvinmetcalf:
I don't know why the tests aren't being shown but they are fully green https://travis-ci.org/nodejs/readable-stream/builds/94369987, :+1:  from anyone in @nodejs/streams ?

sonewman:
Sorry for a delayed response. LGTM :+1:

I was reading through some of the documentation, i think it would be good to revisit some bits, for instance the stream options lists does not include implementation methods.

We could also update the examples to use es2015/6 features now available in Node like classes, shorthand object methods, etc.

Anyway, merging...

calvinmetcalf:
thanks, published as v2.0.5

PR #171, opened 2015-11-26
perf fix for Duplex require replacement

ofrobots:
I was noticing that RegExps called from Module._resolveLookupPaths were getting
to be the hottest symbol in some profiles. readable-stream seems to be requiring
Duplex from inside functions. It would be faster to not do a require on each
invocation of Readable, ReadableState, Writable, WritableState.

### DIFF
--- a/build/files.js
+++ b/build/files.js
@@ -31,9 +31,11 @@ const headRegexp = /(^module.exports = \w+;?)/m
         , '$1\n\n/*<replacement>*/\nvar Buffer = require(\'buffer\').Buffer;\n/*</replacement>*/\n'
       ]
 
+      // The browser build ends up with a circular dependency, so the require is
+      // done lazily, but cached.
     , addDuplexRequire = [
           /^(function (?:Writable|Readable)(?:State)?.*{)/gm
-        , '$1\n  var Duplex = require(\'./_stream_duplex\');\n'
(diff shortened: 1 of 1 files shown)


calvinmetcalf:
This has come up before, there is a circular dependency issue that needs to be resolved

ofrobots:
Is there a test case for the circular dependency issue? With a build from
node 5.1.0, the current tests seem to pass.

calvinmetcalf:
I push up a built version (#172) so we can see how the tests do

calvinmetcalf:
yeah it's an issue with the browser build https://travis-ci.org/nodejs/readable-stream/builds/93369343

ofrobots:
Hmm.. it would nice not pay the extra cost in Node if the just the browser builds are slower. In my simple, but not trivial, express app, I was noticing that the regexp from Module._resolveLookupPaths was taking more than 3% of overall time.

Let me think about if there a different way to do this.

calvinmetcalf:
We could cache the result of the require call so it is only called the
first time

ofrobots:
Done.

ofrobots:
LMK, if I should include the built version in this PR.

calvinmetcalf:
Thanks I'll look at it tomorrow

calvinmetcalf:
tests are running https://travis-ci.org/nodejs/readable-stream/builds/93514719

PR #177, opened 2016-01-18
fix copyright

stevemao:


calvinmetcalf:
it's actually a little more complicated then that https://github.com/nodejs/node/blob/647861befb005004788ea555179c0e962becbb61/LICENSE#L1-L47

stevemao:
Should i copy the whole paragraph?

rvagg:
heads-up that it'll likely change within the coming months, pending https://github.com/nodejs/node/issues/3979, so I wouldn't say there should be a rush to "fix" this cause it was a hack-job in the first place (io.js + convergence removes us 2 steps from the original license)

stevemao:
Makes sense. Closing now.

PR #179, opened 2016-01-21
build for 5.5.0

calvinmetcalf:
initial build of 5.5.0, seams like a lot of changes, lets investigate

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -25,24 +25,22 @@ matrix:
     env: TASK=test
   - node_js: 5
     env: TASK=test
-  - node_js: node
-    env: TASK=test
-  - node_js: node
-    env: TASK=browser BROWSER_NAME=opera BROWSER_VERSION="11..latest"
-  - node_js: node
-    env: TASK=browser BROWSER_NAME=ie BROWSER_VERSION="9..latest"
-  - node_js: node
-    env: TASK=browser BROWSER_NAME=chrome BROWSER_VERSION="41..beta"
-  - node_js: node
-    env: TASK=browser BROWSER_NAME=firefox BROWSER_VERSION="36..latest"
-  - node_js: node
-    env: TASK=browser BROWSER_NAME=ipad BROWSER_VERSION="['6.1', '7.1', '8.2']"
-  - node_js: node
-    env: TASK=browser BROWSER_NAME=iphone BROWSER_VERSION="['6.1', '7.1', '8.2']"
-  - node_js: node
-    env: TASK=browser BROWSER_NAME=safari BROWSER_VERSION="5..latest"
-  - node_js: node
+  - node_js: 5
     env: TASK=browser BROWSER_NAME=android BROWSER_VERSION="4.0..latest"
+  - node_js: 5
+    env: TASK=browser BROWSER_NAME=ie BROWSER_VERSION="9..latest"
+  - node_js: 5
+    env: TASK=browser BROWSER_NAME=opera BROWSER_VERSION="11..latest"
+  - node_js: 5
+    env: TASK=browser BROWSER_NAME=chrome BROWSER_VERSION="-3..latest"
+  - node_js: 5
+    env: TASK=browser BROWSER_NAME=firefox BROWSER_VERSION="-3..latest"
+  - node_js: 5
+    env: TASK=browser BROWSER_NAME=ipad BROWSER_VERSION="6.0..7.1"
+  - node_js: 5
+    env: TASK=browser BROWSER_NAME=iphone BROWSER_VERSION="6.0..7.1"
+  - node_js: 5
+    env: TASK=browser BROWSER_NAME=safari BROWSER_VERSION="5..latest"
 script: "npm run $TASK"
 env:
   global:
--- a/build/build.js
+++ b/build/build.js
@@ -87,7 +87,7 @@ hyperquest(testlisturl).pipe(bl(function (err, data) {
 
   $('table.files .js-directory-link').each(function () {
     var file = $(this).text()
-    if (/^test-stream/.test(file) && !/-wrap\.js$/.test(file))
+    if (/^test-stream/.test(file) && !/-wrap(?:-encoding)?\.js$/.test(file))
       processTestFile(file)
   })
 }))
--- a/build/common-replacements.js
+++ b/build/common-replacements.js
@@ -39,3 +39,8 @@ module.exports.objectKeysReplacement = [
      /Object\.keys/g
    , 'objectKeys'
  ]
+
+module.exports.constReplacement = [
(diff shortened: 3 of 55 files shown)


calvinmetcalf:
ah a lot of these are `var common = require('../common');` -> `require('../common');`

calvinmetcalf:
ok finally green though I want to see if if the iThing bug is fixed by testing against more iThings, overall not much is changed in the actual stream code so we probably don't need to publish after merging this.

calvinmetcalf:
ok rebased would love for a once over by @nodejs/streams despite the wide breadth there is actually very little changed so it's mainly about keeping the build scripts up to date, we probably don't need to publish this as a release.

.travis.yml
```diff
@@ -25,24 +25,22 @@ matrix:
...
-    env: TASK=browser BROWSER_NAME=firefox BROWSER_VERSION="36..latest"
-  - node_js: node
-    env: TASK=browser BROWSER_NAME=ipad BROWSER_VERSION="['6.1', '7.1', '8.2']"
-  - node_js: node
-    env: TASK=browser BROWSER_NAME=iphone BROWSER_VERSION="['6.1', '7.1', '8.2']"
-  - node_js: node
+  - node_js: 5
+    env: TASK=browser BROWSER_NAME=opera BROWSER_VERSION="11..latest"
+  - node_js: 5
+    env: TASK=browser BROWSER_NAME=chrome BROWSER_VERSION="-3..latest"
```
calvinmetcalf: sauce labs was failing on dev (failing to start the browser, tests where not gotten to) so switched to latest, also switched it to do a rolling 3 latest for this and firefox

.travis.yml
```diff
@@ -25,24 +25,22 @@ matrix:
...
+  - node_js: 5
+    env: TASK=browser BROWSER_NAME=firefox BROWSER_VERSION="-3..latest"
+  - node_js: 5
+    env: TASK=browser BROWSER_NAME=ipad BROWSER_VERSION="6.0..7.1"
```
calvinmetcalf: saucelabs now consistently fails to start any 8.0 and higher iThing

build/build.js
```diff
@@ -87,7 +87,7 @@ hyperquest(testlisturl).pipe(bl(function (err, data) {
 
   $('table.files .js-directory-link').each(function () {
     var file = $(this).text()
-    if (/^test-stream/.test(file) && !/-wrap\.js$/.test(file))
+    if (/^test-stream/.test(file) && !/-wrap(?:-encoding)?\.js$/.test(file))
```
calvinmetcalf: there are now 2 steam wrap tests we want to avoid

build/common-replacements.js
```diff
@@ -39,3 +39,8 @@ module.exports.objectKeysReplacement = [
      /Object\.keys/g
    , 'objectKeys'
  ]
+
+module.exports.constReplacement = [
```
calvinmetcalf: there are a lot more const's in the code

build/test-replacements.js
```diff
@@ -138,7 +142,26 @@ module.exports['common.js'] = [
     [
       /var regexp = `\^\(\\\\w\+\)\\\\s\+\\\\s\$\{port\}\/\$\{protocol\}\\\\s`;/,
       `var regexp = '^(\\w+)\\s+\\s' + port + '/' + protocol + '\\s';`
-    ]
+    ],
+    [
+        /require\(['"]stream['"]\)/g
+      , 'require(\'../\')'
+    ],
+    [/forEach\(data, line => \{\n\s+this\.emit\('data', line \+ '\\n'\);\n\s+\}\);/m,
```
calvinmetcalf: there is an arrow function here that I needed to rewrite, if we get too many of them we may want to consider using a transpiler for future builds

build/test-replacements.js
```diff
@@ -214,3 +231,12 @@ module.exports['test-stream-pipe-cleanup-pause.js'] = [
     'var '
   ]
 ]
+module.exports['test-stream2-readable-empty-buffer-no-eof.js'] = [[
+  /let /g,
```
calvinmetcalf: there is a single let, this might actually be a style error in the original test

build/test-replacements.js
```diff
@@ -214,3 +231,12 @@ module.exports['test-stream-pipe-cleanup-pause.js'] = [
...
+  'var '],
+  [
+    `var buf = Buffer(5).fill('x');`,
```
calvinmetcalf: 8 keystrokes were saved in the commit that added this I assume because it looked cool

lib/_stream_readable.js
```diff
@@ -103,6 +103,7 @@ function ReadableState(options, stream) {
   this.needReadable = false;
   this.emittedReadable = false;
   this.readableListening = false;
+  this.resumeScheduled = false;
```
calvinmetcalf: one of 4 actual changes to the stream code

lib/_stream_readable.js
```diff
@@ -725,7 +726,7 @@ Readable.prototype.on = function(ev, fn) {
     this.resume();
   }
 
-  if (ev === 'readable' && this.readable) {
+  if (ev === 'readable' && !this._readableState.endEmitted) {
```
calvinmetcalf: one of 4 actual changes to the stream code

lib/_stream_transform.js
```diff
@@ -80,8 +80,7 @@ function afterTransform(stream, er, data) {
   if (data !== null && data !== undefined)
     stream.push(data);
 
-  if (cb)
-    cb(er);
+  cb(er);
```
calvinmetcalf: one of 4 actual changes to the stream code

lib/_stream_writable.js
```diff
@@ -526,4 +526,5 @@ function endWritable(stream, state, cb) {
       stream.once('finish', cb);
   }
   state.ended = true;
+  stream.writable = false;
```
calvinmetcalf: one of 4 actual changes to the stream code

test/parallel/test-stream2-unpipe-leak.js
```diff
@@ -1,5 +1,5 @@
 'use strict';
-var common = require('../common');
+require('../common');
```
calvinmetcalf: like 45 out of 55 of the files had this change and not much else (numbers arrived at via wild guess)

calvinmetcalf:
if nobody in @nodejs/streams objects I'm going to merge this in latter today so we don't get too far behind

mafintosh:
so meta now. looks good to me. :+1: - thanks for doing this btw

calvinmetcalf:
3119393e50fabbe70b77e0e3c2721b4085a75537

PR #180, opened 2016-01-25
Update docs for a case with a binary safe string reading

TriAnMan:
`readable.setEncoding(null)` - may be the most preferable way to proxy a binary data without any encoding/decoding overhead

### DIFF
--- a/doc/stream.markdown
+++ b/doc/stream.markdown
@@ -414,8 +414,9 @@ hexadecimal string format.
 
 This properly handles multi-byte characters that would otherwise be
 potentially mangled if you simply pulled the Buffers directly and
-called `buf.toString(encoding)` on them.  If you want to read the data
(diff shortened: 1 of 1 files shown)


calvinmetcalf:
wouldn't just using object mode be the preferable way?

TriAnMan:
We had an issue with it.
We have an HTTP server that outs data in chunks of 8196 bytes length (i.e. long json). The content of data is utf8 encoded. If the borders of chunk breaks an utf8 sequence in two parts then each part become converted into an replacement character.
With `readable.setEncoding(null)` - everything is fine and we can combine chunks into a valid data string.

calvinmetcalf:
fryi this is just a copy of the docs, the file you want to update is https://github.com/nodejs/node/blob/master/doc/api/stream.markdown

PR #182, opened 2016-01-26
Identify corresponding node version

jmm:
I was looking for what version of Node the package corresponds to and noticed the outdated `package.json` `description`. Looks like currently you have to check the commit log to see what the corresponding Node version is. What do you think about putting that in the `README` (as PRed here)? That'd require a bit more maintenance effort but would surface the information better as you could see it on npmjs.com / via `$ npm show readable-stream@2.0.5 readme | head` (assuming it doesn't [display outdated `README` content](https://github.com/npm/npm/issues/10652)), and even make it slightly easier to get on GitHub.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,6 +1,6 @@
 # readable-stream
 
-***Node-core streams for userland*** [![Build Status](https://travis-ci.org/nodejs/readable-stream.svg?branch=master)](https://travis-ci.org/nodejs/readable-stream)
+***Node-core v0.0.0 streams for userland*** [![Build Status](https://travis-ci.org/nodejs/readable-stream.svg?branch=master)](https://travis-ci.org/nodejs/readable-stream)
 
 
 [![NPM](https://nodei.co/npm/readable-stream.png?downloads=true&downloadRank=true)](https://nodei.co/npm/readable-stream/)
--- a/build/build.js
+++ b/build/build.js
@@ -6,8 +6,13 @@ const hyperquest  = require('hyperzip')(require('hyperdirect'))
     , path        = require('path')
     , cheerio     = require('cheerio')
     , encoding    = 'utf8'
+    , urlRegex = /^https?:\/\//
     , nodeVersion = process.argv[2]
     , nodeVersionRegexString = '\\d+\\.\\d+\\.\\d+'
+    , usageVersionRegex = RegExp('^' + nodeVersionRegexString + '$')
+    , readmeVersionRegex =
+        RegExp('(Node-core v)' + nodeVersionRegexString, 'g')
+
     , readmePath  = path.join(__dirname, '..', 'README.md')
     , files       = require('./files')
     , testReplace = require('./test-replacements')
@@ -23,14 +28,14 @@ const hyperquest  = require('hyperzip')(require('hyperdirect'))
     , docourroot  = path.join(__dirname, '../doc')
 
 
-if (!RegExp('^' + nodeVersionRegexString + '$').test(nodeVersion)) {
+if (!usageVersionRegex.test(nodeVersion)) {
   console.error('Usage: build.js xx.yy.zz')
   return process.exit(1);
 }
 
 // `inputLoc`: URL or local path.
 function processFile (inputLoc, out, replacements) {
-  var file = /^https?:\/\//.test(inputLoc) ?
+  var file = urlRegex.test(inputLoc) ?
     hyperquest(inputLoc) :
     fs.createReadStream(inputLoc, encoding)
 
@@ -107,8 +112,5 @@ processFile(
 // Update Node version in README
 
 processFile(readmePath, readmePath, [
-  [
-      RegExp('(Node-core v)' + nodeVersionRegexString, 'g')
-    , "$1" + nodeVersion
-  ]
+  [readmeVersionRegex, "$1" + nodeVersion]
 ])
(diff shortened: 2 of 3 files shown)


calvinmetcalf:
The docs do show what version of node it's from

README.md
```diff
@@ -1,6 +1,6 @@
 # readable-stream
 
-***Node-core streams for userland*** [![Build Status](https://travis-ci.org/nodejs/readable-stream.svg?branch=master)](https://travis-ci.org/nodejs/readable-stream)
+***Node-core v5.1.0 streams for userland*** [![Build Status](https://travis-ci.org/nodejs/readable-stream.svg?branch=master)](https://travis-ci.org/nodejs/readable-stream)
```
calvinmetcalf: we probably want to set it so it updates automatically when we run a build

package.json
```diff
@@ -1,7 +1,7 @@
 {
   "name": "readable-stream",
   "version": "2.0.5",
-  "description": "Streams3, a user-land copy of the stream library from iojs v2.x",
+  "description": "Streams3, a user-land copy of the stream library from Node.js",
```
calvinmetcalf: this I am 100% down with right now

jmm:
@calvinmetcalf Thanks for the replies.

> The docs do show what version of node it's from

I feel stupid, but I'm overlooking it, even now. [These docs](https://github.com/nodejs/readable-stream/blob/01fd724f9c997aeb1d10e421268bd99795290ce9/doc/stream.markdown)?

> we probably want to set it so it updates automatically when we run a build

Please take a look at the new commits. I made some general updates to the build script such as:

* Assign `nodeVersion = process.argv[2]` and use that var everywhere else.

* Update the node version regex so it will accept a value like `0.12.9` and reject a value like `v5.1.0`.

* Delete a duplicate test for that.

Then I added updating the `README` on build. If from the root I do `node build/build && npm test` and they pass, that should mean all is well, correct? (I just want to make sure I'm correctly verifying that my changes aren't breaking anything.)

build/build.js
```diff
@@ -5,39 +5,45 @@ const hyperquest  = require('hyperzip')(require('hyperdirect'))
     , fs          = require('fs')
     , path        = require('path')
     , cheerio     = require('cheerio')
-
+    , encoding    = 'utf8'
+    , nodeVersion = process.argv[2]
+    , nodeVersionRegexString = '\\d+\\.\\d+\\.\\d+'
```
calvinmetcalf: just make this `/^\\d+\\.\\d+\\.\\d+$/` no reason to create a string latter instead of a regexp now

build/build.js
```diff
@@ -103,3 +102,13 @@ processFile(
   , path.join(testourroot, '../common.js')
   , testReplace['common.js']
 )
+
+//--------------------------------------------------------------------
+// Update Node version in README
+
+processFile(readmePath, readmePath, [
+  [
+      RegExp('(Node-core v)' + nodeVersionRegexString, 'g')
```
calvinmetcalf: you can reuse the same regex you used to test above and have it just [versionRegex, nodeVersion]

calvinmetcalf:
looks great, I think we can simplify it but it's otherwise great

jmm:
@calvinmetcalf Thanks for the feedback. I don't see any way around needing at least 2 regexes (whether they're defined literally or otherwise).

> just make this `/^\\d+\\.\\d+\\.\\d+$/` no reason to create a string latter instead of a regexp now

> you can reuse the same regex you used to test above and have it just [versionRegex, nodeVersion]

If there's one anchored regex it'll work for the usage test but not for the `README` update test. If there's one unanchored regex, the usage test will allow input that won't really work, like `v5.1.0`, and if it's global, to account for more than one instance in the `README` that should be auto-updated, it'll also update stuff like:

> As of version 2.0.0 **readable-stream**

Your first suggestion could be combined with using another regex for the `README` test, like:

```js
[
  /(Node-core v)([^ ]+)/g,
  function (match, p1, p2) {
    if (p2.test(nodeVersionRegex) match = p1 + nodeVersion;
    return match;
  }
]
```

Or just assume anything after the `v` is a semver string. Or for more flexibility / explicitness mark it up in the `README` with something like `Node-core v<span class="auto-update-node-version">0.0.0</span>` and use a regex to match that and not even worry about matching the semver part. Either way there's more than one regex though.

calvinmetcalf:
ah I see, since they're going to be the same every time lets define them up top, the https one as well

jmm:
@calvinmetcalf Done :sparkles:. FWIW there are other pre-existing regexes defined inline where they're used, e.g.:

https://github.com/nodejs/readable-stream/blob/01fd724f9c997aeb1d10e421268bd99795290ce9/build/build.js#L90

https://github.com/nodejs/readable-stream/blob/01fd724f9c997aeb1d10e421268bd99795290ce9/build/build.js#L102

calvinmetcalf:
yeah there is some pretty gnarly code, I'm just trying not to make it too much work :smile:

jmm:
Thanks @calvinmetcalf.

In hindsight in the end I should've left the `README` with the current version, instead of `v0.0.0` (I just had it that way so that I could just keep running the build after each change to make sure the `README` updated, then check it out again.) But it's really no worse than it was before and it'll update whenever the build runs (if it were run with `build 5.1.0` right now just the README would update).

PR #185, opened 2016-02-16
build for 5.6.0

calvinmetcalf:


calvinmetcalf:
and we are actually green! https://travis-ci.org/nodejs/readable-stream/builds/109629800

calvinmetcalf:
can I get a :+1: on this from anyone in @nodejs/streams so I can start work on v5.7.0?

mcollina:
LGTM, but I'm not in @nodejs/streams (how can I join?)

calvinmetcalf:
that is a very good question looks like @chrisdickinson  or @rvagg would have to invite you

rvagg:
Done, but I think if you click through to @nodejs/streams you can request membership at least.

@calvinmetcalf could do with a hand around here I think! Thanks @mcollina.

mcollina:
Thanks @rvagg! It's not clear _how_ you join a WG for the docs.

rvagg:
that's up to the WG to figure out, some are more open than others by nature

mcollina:
@rvagg of course! :) But even that is not written anywhere.

PR #186, opened 2016-02-24
build based on node 5.7.0

calvinmetcalf:
this will probably end up in a readable-stream release


calvinmetcalf:
all the one off fixes for arrow functions are pretty untenable, so this version uses babel, which adds 2 wrinkles

- some of the replacements need to be rewritten to use const or let because now those are being fixed after the replacements
- a couple tests have top level returns so need to be wrapped in an iefe (which is why babel must go after the replacements).

babel does something funky with the white space so while I have retainLines on right now to be able to eye ball the changes we probably want to turn that off.

calvinmetcalf:
and we are green https://travis-ci.org/nodejs/readable-stream/builds/111795595

mcollina:
what's the output without retainLine?

I would prefer different line numbers from core but a more readable thing. Line numbers can potentially be different anyway (because of the replacements).

calvinmetcalf:
It puts conditionals that don't have curly braces are put on to a single
line, not a big deal but it makes comparing the diff tricky, I can rerun it
with that off now that it passes.

calvinmetcalf:
ok removed retain lines

mcollina:
:+1:. I would bump the minor here, just because it's a significant change (babel), and my change might or might not land in v4.x.x.

calvinmetcalf:
hm I tend to not count build changes in terms of features, but it is a
largish change so I'm on the fence here

lib/_stream_readable.js
```diff
@@ -484,15 +441,12 @@ function maybeReadMore(stream, state) {
...
-         state.length < state.highWaterMark) {
+  while (!state.reading && !state.flowing && !state.ended && state.length < state.highWaterMark) {
     debug('maybeReadMore read 0');
     stream.read(0);
     if (len === state.length)
       // didn't get any data, stop spinning.
-      break;
-    else
-      len = state.length;
+      break;else len = state.length;
```
mafintosh: this looks weird

calvinmetcalf: i agree, but it is an artifact of babel transforming the code

calvinmetcalf:
hey guys can I get a yay or nay on this one

mcollina:
LGTM

PR #189, opened 2016-03-12
build for 5.8.0

calvinmetcalf:
there are really no changes here so may as well merge this one before publishing the next version


mcollina:
LGTM
Il giorno sab 12 mar 2016 alle 14:42 Calvin Metcalf <

calvinmetcalf:
ok there was a failure , but it looked like it only showed up when you had 2 builds running on top of each other (aka one build starts before the other one ends, it shows up in the 2nd one) so it's probably a timing issue.

PR #191, opened 2016-03-29
build for 5.9.1

calvinmetcalf:
this is one is pretty strait forward


mcollina:
:+1:

calvinmetcalf:
wait knew that was too easy, github changed the classes in the table so our script broke, that being said, nothing actually changed

mcollina:
even better :).

phillipj:
Woopsie, ignore that comment... Some testing going on here :)

PR #197, opened 2016-04-11
add enviromental variable which causes node streams to be passed through/testing improvments

calvinmetcalf:
also adds browserify transforms to prevent this from ballooning the size

### DIFF
--- a/package.json
+++ b/package.json
@@ -6,20 +6,26 @@
   "dependencies": {
     "core-util-is": "~1.0.0",
     "inherits": "~2.0.1",
+    "inline-process-browser": "~2.0.1",
     "isarray": "~1.0.0",
     "process-nextick-args": "~1.0.6",
     "string_decoder": "~0.10.x",
+    "unreachable-branch-transform": "~0.5.0",
     "util-deprecate": "~1.0.1"
   },
   "devDependencies": {
-    "tap": "~0.2.6",
+    "nyc": "^6.4.0",
+    "tap": "~0.7.1",
     "tape": "~4.5.1",
     "zuul": "~3.9.0"
   },
   "scripts": {
     "test": "tap test/parallel/*.js test/ours/*.js",
     "browser": "npm run write-zuul && zuul -- test/browser.js",
-    "write-zuul": "printf \"ui: tape\nbrowsers:\n  - name: $BROWSER_NAME\n    version: $BROWSER_VERSION\n\">.zuul.yml"
+    "write-zuul": "printf \"ui: tape\nbrowsers:\n  - name: $BROWSER_NAME\n    version: $BROWSER_VERSION\n\">.zuul.yml",
+    "local": "zuul --local -- test/browser.js",
+    "cover": "nyc npm test",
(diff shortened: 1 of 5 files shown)


mcollina:
Why `envify` and `unreacheable-branch-transform`? Using `stream` in the browser will lead to different results? Is this truly needed?

BTW, we'll need to backport this to the 1.x line as well, just in case.

phillipj:
@calvinmetcalf fyi the Travis build took to long, so the bot gave up polling Travis for build status. That means the inline Travis CI status in this PR will be stale.

We should probably increase how long the bot tries before giving up.

calvinmetcalf:
@mcollina that is so that this will not increase the size of the browserify build because without those then this change would cause streams to always be included with readable streams even if never required anywhere else

@phillipj yeah I think it can take 3 hours or more

mcollina:
@calvinmetcalf put a little comment in the source, pointing it to those two modules, for the reader who does not know about them.

Is this related somehow to?
https://github.com/nodejs/readable-stream/blob/master/readable.js#L1-L5

calvinmetcalf:
actually since streams in browserify are based on readable stream we probably want to turn this off in the browser and use inline-process-browser instead

calvinmetcalf:
ok also added the ability to run the browser tests locally which allows you to see code coverage of the browser tests

mcollina:
LGTM

calvinmetcalf:
and got working coverage

calvinmetcalf:
ugh I'd kind of like this to be green before I merge but all of the tests have been yet to be green at the same time (Though it's different ones getting the zuul bug)

package.json
```diff
@@ -6,20 +6,26 @@
   "dependencies": {
     "core-util-is": "~1.0.0",
     "inherits": "~2.0.1",
+    "inline-process-browser": "~2.0.1",
     "isarray": "~1.0.0",
     "process-nextick-args": "~1.0.6",
     "string_decoder": "~0.10.x",
+    "unreachable-branch-transform": "~0.5.0",
```
sindresorhus: Shouldn't `inline-process-browser` and `unreachable-branch-transform` be in `devDependencies`?

These dependencies add 2 MB to the package.

mcollina: No, they are used to avoid bundling unneeded stuff for the browserify build.
Are you saying that these adds up in the browserify build?

sindresorhus: No, they add up for Node.js modules depending on `readable-stream`. 2 MB extra is a *lot* for small modules.

mcollina: We need a way to force "readable-stream" to use node own stream module. We need this to verify new streams implementations do not break existing modules before releasing new versions of node.
This feature increases the browserify bundle by a huge factor, so the two transform for browserify were included.
I'll be more than happy to have an alternative implementation that achieve both goals.

sindresorhus: I don't have any immediate alternative, but I've opened an issue so it can be tracked: https://github.com/nodejs/readable-stream/issues/205

PR #199, opened 2016-04-13
backport env variable change

calvinmetcalf:
for users on the old ~1.0.0 way of doing it, no need for fancy browserify stuff here


calvinmetcalf:
yeah this will never pass travis, the idea will be to publish this without the latest tag as 1.0.34

mcollina:
agreed, lgtm
Il giorno mer 13 apr 2016 alle 17:37 Calvin Metcalf <

PR #200, opened 2016-04-13
backport env fix

calvinmetcalf:
for those on the ~1.1.0 version


calvinmetcalf:
this will never pass travis, the idea is to publish this without the latest tag as 1.1.14

calvinmetcalf:
specifically I'll use the 1.1 tag that was previously used

calvinmetcalf:
actually tagged as backport-1.1.x because they tightened things up and I can no longer tag as 1.1

PR #206, opened 2016-04-27
reuse existing import of "stream"

### LINKED ISSUE #205: Reduce package size
From https://github.com/nodejs/readable-stream/pull/197#discussion_r61090554.

Having to download 2 MB extra makes this module unusable for most smaller modules.

Would be nice to find a way to not have to include those huge dependencies.

calvinmetcalf:
should fix #205

### DIFF
--- a/package.json
+++ b/package.json
@@ -6,11 +6,9 @@
   "dependencies": {
     "core-util-is": "~1.0.0",
     "inherits": "~2.0.1",
-    "inline-process-browser": "~2.0.1",
     "isarray": "~1.0.0",
     "process-nextick-args": "~1.0.6",
     "string_decoder": "~0.10.x",
-    "unreachable-branch-transform": "~0.5.0",
     "util-deprecate": "~1.0.1"
   },
   "devDependencies": {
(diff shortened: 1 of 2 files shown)


mcollina:
LGTM. BTW, what's the status on our CITGM with this?

calvinmetcalf:
ok will merge and publish, as for citgm no idea was in japan and dubai last week and was sick as a dog most of this week so I'm totally out of the loop.

ahutchings:
This breaks browserify builds because the removed dependencies are still listed in package.json under browserify transforms. Should the browserify configuration be removed from package.json as well?

calvinmetcalf:
fuck your right one sec

calvinmetcalf:
thanks for the catch @ahutchings, unclear how that passed our tests which involve browserify...

ahutchings:
No worries, thanks for the quick turnaround.

calvinmetcalf:
np

calvinmetcalf:
ok I'm going to depreciate 2.1.1 with the following message

  Version 2.1.1 has a misconfigured package.json and will likely not work with browserify, If you aren't directly requiring readable-stream use `npm ls readable-stream@2.1.1` to find out what package is.

since this is reversible I'm going to just go ahead and do this.

Not going to unpublish since from experience that causes a lot of problems, but we can if we want latter

PR #208, opened 2016-05-06
build for v6.1.0

calvinmetcalf:
seems like the big changes are the new buffer api, plus that fix of mine for throwing on null chunks


lib/_stream_readable.js
```diff
@@ -149,7 +162,7 @@ Readable.prototype.push = function (chunk, encoding) {
   if (!state.objectMode && typeof chunk === 'string') {
     encoding = encoding || state.defaultEncoding;
     if (encoding !== state.encoding) {
-      chunk = new Buffer(chunk, encoding);
+      chunk = Buffer.from(chunk, encoding);
```
mafintosh: this doesn't work on older nodes right?

lib/_stream_readable.js
```diff
@@ -825,16 +839,16 @@ function fromList(n, state) {
     } else {
       // complex case.
       // we have enough to cover it, but it spans past the first buffer.
-      if (stringMode) ret = '';else ret = new Buffer(n);
+      if (stringMode) ret = '';else ret = Buffer.allocUnsafe(n);
```
mafintosh: same as above

lib/_stream_writable.js
```diff
@@ -265,11 +271,12 @@ Writable.prototype.setDefaultEncoding = function setDefaultEncoding(encoding) {
   if (typeof encoding === 'string') encoding = encoding.toLowerCase();
   if (!(['hex', 'utf8', 'utf-8', 'ascii', 'binary', 'base64', 'ucs2', 'ucs-2', 'utf16le', 'utf-16le', 'raw'].indexOf((encoding + '').toLowerCase()) > -1)) throw new TypeError('Unknown encoding: ' + encoding);
   this._writableState.defaultEncoding = encoding;
+  return this;
 };
 
 function decodeChunk(state, chunk, encoding) {
   if (!state.objectMode && state.decodeStrings !== false && typeof chunk === 'string') {
-    chunk = new Buffer(chunk, encoding);
+    chunk = Buffer.from(chunk, encoding);
```
mafintosh: same as above

lib/_stream_readable.js
calvinmetcalf: yeah we're probably going to have to write a shim or polyfill

calvinmetcalf:
next up will be a fix for deepStrictEqual

calvinmetcalf:
defunctzombie/commonjs-assert#17

Fishrock123:
@calvinmetcalf any progress? Need any help?

calvinmetcalf:
I need to port deepStrictEqual to work on all the platforms defunctzombie/commonjs-assert#17 is not going as well as I'd hoped

calvinmetcalf:
update looked like I found the problem, it was ArrayBuffer.isView

calvinmetcalf:
and I should have been suspicious about all the file changes

calvinmetcalf:
ok while we wait for @defunctzombie to review the changes to assert-commonjs we can test against my branch to make sure it actually fixes things

defunctzombie:
@calvinmetcalf I've added you as collaborator to the commonjs-assert repo. I don't have much time for some of these projects these days and don't want to be a blocker.

calvinmetcalf:
thanks!

calvinmetcalf:
and got it working with defunctzombie/commonjs-assert#18

calvinmetcalf:
ok I've had all browsers pass, just not in the same run so screw it calling that green and going to publish

PR #209, opened 2016-05-06
fixed the function call for asyncWrite

stockulus:
Hi,

the "asyncWrite" is actually a process.NextTick or setImmediate. So the usage in the async after write process is wrong in my opinion, and causes a Problem for me (currently working on a React native Project).

If you need further information, just contact me.
@stockulus on twitter or stockulus@icloud.com  

Thanks.
Christoph

### DIFF
--- a/lib/_stream_writable.js
+++ b/lib/_stream_writable.js
@@ -346,7 +346,9 @@ function onwrite(stream, er) {
 
     if (sync) {
       /*<replacement>*/
-      asyncWrite(afterWrite, stream, state, finished, cb);
+      asyncWrite(function() {
+        afterWrite(stream, state, finished, cb);
+      });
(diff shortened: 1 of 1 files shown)


calvinmetcalf:
so this is built from what's in nodejs so you'd want to open a pull there to fix it (specifically [this line](https://github.com/nodejs/node/blob/98b534ff5ae67c6381414126cd6f78d53f4cc932/lib/_stream_writable.js#L350)).

That being said you're probably going to go into a little more detail why you thing it's wrong and why you think nesting 2 asyncwrites makes it better ?

stockulus:
sorry for you time, this was my fault, polyfill react-native not correctly, so the process-args did not work correctly.

Issue #212, opened 2016-05-19
Remove isarray dependency 


jimmywarting:
think we can replace it with `Array.isArray`

calvinmetcalf:
Not on all targeted platforms I don't believe

mcollina:
Is `Array.isArray` present everywhere on all our target platforms?

jimmywarting:
Wich platform dö we target?

calvinmetcalf:
actually now that I look at it, isArray seems to be available on all the platforms, we can test

dcousineau:
The `isarray` package in its entirety is:

```javascript
var toString = {}.toString;

module.exports = Array.isArray || function (arr) {
  return toString.call(arr) == '[object Array]';
};
```

And can more simply be inlined, not required.

Croydon:
Before I have opened #318 I should be searched first. Shame on me :) 

So what's the plan here? You discussed in the pull request which was going to remove the isArray dependency, then you decided to make it inline https://github.com/nodejs/readable-stream/commit/69d13e3747f7c3596a6cbaf266f43f6d8e5fcb7d#diff-279988bef5e6fbf7ed82cae28ecc2d59 just to revert this again in the next commit https://github.com/nodejs/readable-stream/commit/ea4eabaa30929f685ec1c3955587ec4513b31ae8#diff-279988bef5e6fbf7ed82cae28ecc2d59

So...what is the plan?

jimmywarting:
Bumping my thread 2 years later after still seeing it being used in my dependency graph from karma...

Can we bump the major version soon and remove the dependency and perhaps the inlined version also?

looks pretty green to me:
![skarmavbild 2018-04-27 kl 20 03 32](https://user-images.githubusercontent.com/1148376/39377645-33ec66d0-4a56-11e8-9ed7-1d1e451fa63b.png)

mcollina:
Yes, we will do that in the coming weeks. We will cut a new major release from Node 10 as soon as possible (but it will take a while).

PR #214, opened 2016-05-19
testing to see if Array.isArray really is required

calvinmetcalf:
#212

### DIFF
--- a/build/files.js
+++ b/build/files.js
@@ -70,16 +70,6 @@ const headRegexp = /(^module.exports = \w+;?)/m
         + '}catch(_){}}());\n'
       ]
 
-    , isArrayDefine = [
-          headRegexp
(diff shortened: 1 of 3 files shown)


mcollina:
It seems fine, LGTM.

calvinmetcalf:
I am somewhat hesitant just that there may be some platforms that we can't test on that it does work enough on but would break if we remove this

mcollina:
Let's stick a semver-major label, and we'll merge when we bump the major.
Il giorno lun 23 mag 2016 alle 13:44 Calvin Metcalf <

calvinmetcalf:
but why should we bother, it's like 3 lines, there is very little upside

mafintosh:
I agree with @calvinmetcalf that I dont see the upside in merging this

Issue #217, opened 2016-06-03
Introduction of ngrok introducing failures on ppc + osx in citgm


MylesBorins:
Looks like the ngrok download is failing on osx. It also looks like ngrok does not have a ppc binary, which is causing failures on that architecture.

it seems like this dep was added in 2.1.4

Wanted to see what ya'll thought before I add readable stream to the flaky list for those platforms

mcollina:
I think we can _keep_ using ngrok, but install only on Travis. That would simplify things for citgm.

calvinmetcalf:
as it turns out ngrok isn't working for some other reasons

mcollina:
This should be fixed.

PR #218, opened 2016-06-09
6.2.1

calvinmetcalf:
what I like to see, only tests added


mcollina:
Tests failing :( some of them because of Error: ENOENT: no such file or directory, open '/home/travis/build/nodejs/readable-stream/test/fixtures/keys/ec-key.pem'

calvinmetcalf:
we don't actually need that file, it's for the c api

lpinca:
browser failures are probably caused by ngrok limits on free plan. Also there is bug in the ngrok module that prevents open tunnels from closed, see https://github.com/defunctzombie/zuul/issues/251#issuecomment-223586880.

In light of #217 it might makes sense to switch back to localtunnel (it worked fine for me in the last few days.)

calvinmetcalf:
ok trying without ngrok

mcollina:
@calvinmetcalf why don't we ask ngrok if they could sponsor an account for CI to the Foundation? Or maybe we can buy one. 

Another solution is to setup our own localtunnel server, I think we should try to avoid that as far as possible (but it's a solution).

lpinca:
Another option is to use karma + karma-sauce-launcher instead of zuul but the ppc issue will not be solved. I don't think there is a ppc binary for sauce-connect.

lpinca:
I think zuul can also run with sauce-connect.

calvinmetcalf:
@lpinca thanks will try that

calvinmetcalf:
woot that works thanks @lpinca that being said there is a suble timing bug in one of the tests in 0.8 that just for fun, can be prevented by putting certain console.log statments in place

calvinmetcalf:
ok from what I can tell, in node 0.8 setTimeout(fn, 0) will sometimes comeback before process.nextTick

calvinmetcalf:
more specifically I think setTimeout(fn, 0) is the same as process.nextTick, making it setTimeout(fn, 1) fixes things

calvinmetcalf:
well 1 doesn't but 4 does

mcollina:
I think we should drop 0.8 asap and bump the major version.

calvinmetcalf:
naw I got 0.8 working

package.json
```diff
@@ -17,13 +17,12 @@
...
     "tape": "~4.5.1",
-    "zuul": "~3.10.0",
-    "zuul-ngrok": "^4.0.0"
+    "zuul": "~3.10.0"
   },
   "scripts": {
     "test": "tap test/parallel/*.js test/ours/*.js",
     "browser": "npm run write-zuul && zuul -- test/browser.js",
-    "write-zuul": "printf \"ui: tape\ntunnel: ngrok\nbrowsers:\n  - name: $BROWSER_NAME\n    version: $BROWSER_VERSION\n\">.zuul.yml",
+    "write-zuul": "printf \"ui: tape\nsauce_connec: true\nbrowsers:\n  - name: $BROWSER_NAME\n    version: $BROWSER_VERSION\n\">.zuul.yml",
```
lpinca: There is a typo here: `sauce_connec` → `sauce_connect`.

calvinmetcalf: fixed thanks

sonewman:
hmm, there does seem to still be an issue :(

lpinca:
@sonewman if you are talking about CI, yes zuul is flaky as hell and `sauce_connect` make it even more flaky :/

As @mcollina said, it would be great if we can get a ngrok account.

calvinmetcalf:
**sign** yeah I think your right

mcollina:
We should probably update this to 6.2.2 while we are at it https://github.com/nodejs/readable-stream/issues/220.

calvinmetcalf:
ok added #220

calvinmetcalf:
though in this case it was because I a new es6 feature was introduced

calvinmetcalf:
bumped it, i don't think there is anything actually wrong with the pull, just the tests

mcollina:
There is this change: https://github.com/nodejs/readable-stream/pull/218/files#diff-ba6a0df0f5212f5cba5ca5179e209a17L544.

calvinmetcalf:
yeah I take this back a browser is failing legitimately

calvinmetcalf:
testing manually to see if I can reproduce the android error

calvinmetcalf:
weird not able to reproduce the android error when running saucelabs locally

lpinca:
@calvinmetcalf I didn't have issues with localtunnel lately, maybe try to switch again? 😄

calvinmetcalf:
ok all browsers have had at least one green run so I'm calling this green enough

PR #219, opened 2016-06-15
Pass Number.MAX_SAFE_INTEGER to stream.read to improve performance

ajschlosser:
I discovered this issue while working on a totally different project that uses pdfmake.

See bpampuch/pdfmake/issues/280 for more details.

By passing `Number.MAX_SAFE_INTEGER` to `stream.read` in the `flow()` function, performance improved by several orders of magnitude for really, really large amounts of data.

Thoughts?

### DIFF
--- a/lib/_stream_readable.js
+++ b/lib/_stream_readable.js
@@ -737,7 +737,7 @@ function flow(stream) {
   debug('flow', state.flowing);
   if (state.flowing) {
     do {
-      var chunk = stream.read();
+      var chunk = stream.read(Number.MAX_SAFE_INTEGER || 9007199254740991);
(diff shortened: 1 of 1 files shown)


calvinmetcalf:
So first this is generated out of the nodejs repo so you'd want to open a
pull though there, second I'd need to check the code you're using but using
the highWaterMark option might do what you need

calvinmetcalf:
ok commenting on the linked issue, but tl;dr you're code that's consuming the stream almost definitely has the performance issue

ajschlosser:
Appreciate the quick response. Thanks for commenting in the linked issue, too. I'll implement your suggestions(s) and see if that gets to the root of the problem.

PR #223, opened 2016-07-25
handle arbitrary EE implementations

KoryNunn:
Currently, if an emitter that does not inherit from EventEmitter is passed, an error will be thrown:

`Error: emitter.prependListener is not a function`

This PR checks for the existence of prependListener on the emitter that is passed in.

### DIFF
--- a/lib/_stream_readable.js
+++ b/lib/_stream_readable.js
@@ -55,10 +55,8 @@ var StringDecoder;
 
 util.inherits(Readable, Stream);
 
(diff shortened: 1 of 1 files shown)


calvinmetcalf:
this is actually a seriese of regexes that transform the node version, if you can just add a test to in [this](https://github.com/nodejs/readable-stream/tree/master/test/ours) folder I can work on a fix

calvinmetcalf:
see nodejs/node#8018

PR #225, opened 2016-08-17
first run at 6.4.0

calvinmetcalf:
and there is a new es6 feature being used in tests and a couple more fixtures!


calvinmetcalf:
and a few tests that use encodings not all versions of node have, no actual changes to streams here so no rush

calvinmetcalf:
ok the issue with the preprocessing test was basically it was relying on some specific tty stuff that was kinda tangential to streams specifically

calvinmetcalf:
if others are on the same page about being able to ditch that preprocess test I can merge this

mcollina:
@calvinmetcalf LGTM

As a side note, that test is not targeting `readable-stream`, as it is using `fs`, `readline` and others`, so there is nothing we can fix here.

Can we avoid pulling it the next time? That test also relies on two `txt` files, those should be removed as well. It's really a test about `fs` and `readline` rather than streams.

As a side note, that test it was introduced by @vsemozhetbyt in https://github.com/nodejs/node/pull/7741 and https://github.com/nodejs/node/commit/ea725eda426add1d684c80f2c2c585cd2ca797c9

cc @addaleax

calvinmetcalf:
oh right need to remove those

calvinmetcalf:
we'd avoid pulling it in if it wasn't named 'stream' it's just a function of the fact that it was touching 3 diff modules and it had to pick one

mcollina:
@calvinmetcalf we should move it then.

addaleax:
I’m not sure what the process for pulling in tests is but if moving that one test in the node repo makes anything easier for anyone that shouldn’t be a problem?

calvinmetcalf:
not a big deal any more, I just put something in the code that doesn't pull
it in

calvinmetcalf:
this is a transient error, I'll plan on merging this latter if nobody objects

PR #227, opened 2016-08-29
Fix typo in documentation

hubertmine:


calvinmetcalf:
these are copied from the nodejs repo so you'd want to update it there

hubertmine:
Ok thanks, the PR is already opened.

PR #230, opened 2016-09-06
first stab at 6.5.0

calvinmetcalf:
should be pretty straight forward as the only code difference seems to be a commit that I already incorporated into this lib


calvinmetcalf:
considering no code changed this is obviously a transitory bug, going to merge

PR #234, opened 2016-09-23
Remove isarray dep.

jdalton:
I know it's been brought up a few times (#212) but I think it's time again.

The [isarray](https://www.npmjs.com/package/isarray) package [is being deprecated](https://github.com/juliangruber/isarray/issues/10) since IE < 11 is no longer supported by MS & IE 8 was the last browser that needed it.

The `Array.isArray` method is supported by all Node versions, Rhino, iOS, Android, and headless browser enviros (like PhantomJS v1, SlimerJS).

### DIFF
--- a/build/files.js
+++ b/build/files.js
@@ -70,16 +70,6 @@ const headRegexp = /(^module.exports = \w+;?)/m
         + '}catch(_){}}());\n'
       ]
 
-    , isArrayDefine = [
-          headRegexp
-        , '$1\n\n/*<replacement>*/\nvar isArray = require(\'isarray\');\n/*</replacement>*/\n'
-      ]
-
-    , isArrayReplacement = [
-          /Array\.isArray/g
-        , 'isArray'
-      ]
-
     , objectKeysDefine = require('./common-replacements').objectKeysDefine
 
     , objectKeysReplacement = require('./common-replacements').objectKeysReplacement
@@ -220,8 +210,6 @@ module.exports['_stream_readable.js'] = [
   , altIndexOfUseReplacement
   , instanceofReplacement
   , stringDecoderReplacement
-  , isArrayDefine
-  , isArrayReplacement
   , debugLogReplacement
   , utilReplacement
   , stringDecoderReplacement
--- a/lib/_stream_readable.js
+++ b/lib/_stream_readable.js
@@ -6,10 +6,6 @@ module.exports = Readable;
 var processNextTick = require('process-nextick-args');
 /*</replacement>*/
 
-/*<replacement>*/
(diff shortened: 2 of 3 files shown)


calvinmetcalf:
yeah it's probably time, do you mind reverting the changes to `lib/_stream_readable.js` ? the next time we build the changes will get incorporated in

jdalton:
> do you mind reverting the changes to lib/_stream_readable.js ? 

All the changes to `lib/_stream_readable.js`? Won't that break things? I based the PR on https://github.com/nodejs/readable-stream/pull/214/files

calvinmetcalf:
eh good point, so @nodejs/streams do we call this a major version as it's strictly speaking not backwards compat? or do we call it an internal api change that doesn't need a major version bump

calvinmetcalf:
ah yes, now I remember why I didn't want to remove it before

jdalton:
> This is semver-major for me.

Big IE8 concern?

calvinmetcalf:
we actually probably [don't work](https://github.com/nodejs/readable-stream/blob/7ca6e8e867fc996866d651db0a6b12fd240bb06e/lib/_stream_readable.js#L785) in IE8 at the moment anyway so this might not technically be a semver major, because we already did the semver major thing and nobody noticed

mcollina:
That's why we should have a conversation about versioning.  At least it is clear to us and to the community.

kgryte:
@calvinmetcalf A distinction exists between intent and actuality. If using `isarray` was intended to address backward compatibility concerns, the spirit of the package would be to work in older environments, regardless of whether this is the case in practice. Hence, removing the dependency (without inlining) should be a semver major, as the intent of the package would be to not support older environments (the intent would be to break compatibility). That, by defn, would be a semver major.

But, as I stated before, I think the intent to drop backward compatibility is unwarranted.

jdalton:
@kgryte As @calvinmetcalf [mentioned](https://github.com/nodejs/readable-stream/pull/234#issuecomment-249283907) there is little to no back compat concern. This is just a case of removing an superfluous dep.

mcollina:
Even if I agree with you, I still think we should talk about the versioning stragegy of this one, currently there is none.

jdalton:
Ironing out versioning strategies is a win 👍

calvinmetcalf:
we could inline it as a short term solution to avoid the dep which people seem to really not like, while we discus versioning

jdalton:
No need for short term fixes. A major version bump is fine.

calvinmetcalf:
at the wg meeting (#235) we decided on inlining (#238) as having all the benefit with minimal (even theoretical) downside

PR #239, opened 2016-10-15
docs: provide links to stream types in streams.md

lrlna:
Hi folks! 

I was spending some time reading through the documentation for readable-stream, and I found that a few anchor tags weren't getting directed to the right places. This PR hooks up stream types to their respective classes. Do let me know if the intention was to do something else, and I will change it. 

P.S. I couldn't find a `contributing.md`, so just let me know if I need to change anything in the way this pull request was created as well ^_^

### DIFF
--- a/doc/stream.md
+++ b/doc/stream.md
@@ -35,13 +35,13 @@ the elements of the API that are required to *implement* new types of streams.
 
 There are four fundamental stream types within Node.js:
 
-* [Readable][] - streams from which data can be read (for example
+* [Readable](#class-streamreadable) - streams from which data can be read (for example
   [`fs.createReadStream()`][]).
-* [Writable][] - streams to which data can be written (for example
+* [Writable](#class-streamwritable) - streams to which data can be written (for example
   [`fs.createWriteStream()`][]).
 * [Duplex](#class-streamduplex) - streams that are both Readable and Writable (for example
   [`net.Socket`][]).
-* [Transform][] - Duplex streams that can modify or transform the data as it
+* [Transform](#class-streamtransform) - Duplex streams that can modify or transform the data as it
   is written and read (for example [`zlib.createDeflate()`][]).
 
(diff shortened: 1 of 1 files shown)


yoshuawuyts:
Woah nice! LGTM :tada:

mafintosh:
Looks great! Just double checking with @calvinmetcalf that we don't pull in these docs from node core

lrlna:
@mafintosh oooooo that would actually make sense; I just double checked in node core, and they are identical with what's here. I suppose it would be better to make this pull there?

vweevers:
I don't think you can fix this in node core. These links [won't work on GitHub](https://github.com/nodejs/node/blob/master/doc/api/stream.md) (or any markdown renderer), because the Node.js doc generator [prefixes all anchor urls](https://github.com/nodejs/node/blob/fcee4d4ad9cb44971ad9ba2a5583285fc34c7708/tools/doc/html.js#L343) with the basename of the document. That's why a link references  `#stream_class_stream_readable` rather than `#class_stream_readable`.

calvinmetcalf:
Sorry about to go on vacation, these do come from node core but we may need
to change the script to fix links to work with github

mcollina:
@lrlna we need to fix this in https://github.com/nodejs/readable-stream/blob/master/build/doc-replacements.js and https://github.com/nodejs/readable-stream/blob/master/build/build.js, not in the README itself

lrlna:
okay, gotcha! @mcollina do y'all prefer to continue to pull the correct info from core, or would you prefer me to fix the build here first? 

And, I mean, it's broken in core, so I might as well just do it there \o/

mcollina:
I think the best path is to fix them in core, and then check if we need to update our build scripts.

lrlna:
@mcollina okay! I'll close this since it's no longer relevant and send a pr to core; stay tuned for the fix.

PR #241, opened 2016-10-25
npm package: exclude docs, fixes #240

### LINKED ISSUE #240: npm package is half doc/stream.md
The readable-stream NPM package, when installed, is 129k.  Half of that size is stream.md, currently 66k.  Why is this (large) doc distributed with the NPM package?  Would you accept a PR to remove it (add doc/ to .npmignore)?

yzorg:
Fixes #240.  Save 60k per install.

### DIFF
--- a/.npmignore
+++ b/.npmignore
@@ -1,4 +1,5 @@
 build/
+doc/
 test/
 examples/
 fs.js


mcollina:
I'm generally against this type of changes (I check docs, read tests when I am offline), but we already have a `.npmignore` in this project, so it's ok.

calvinmetcalf:
the docs are intentionally in there so people can read them offline

yzorg:
@calvinmetcalf There are three copies of this package installed when installing the current gulp package (via transitive dependencies, and stale intermediate packages).  The folks who need the docs can easily clone the GIT repo, right?  

While this package is a clone of APIs that ship with node 6.5, gulp is still actively supporting node 0.10, so they will probably never remove this package dependency from the current major version (gulp 3.x).  And, I would be with you if we were talking about 100k installs per month, but gulp is downloaded from NPM 2.6 million times last month.

mcollina:
@yzorg it's 20KB of downloads, and 60KB on disk. 
42.148.038 times 20KB is a big number, around 800GB.
Given that everyone depends on this module, it might make sense to do it.
Why don't we talk about it in the WG meeting?

mcollina:
As discussed during the wg meeting, it seems we are ok in removing docs.

Any opinion against?

calvinmetcalf:
if we don't include them we should probably just stop generating them as well, the whole point was that you could read them offline, especially as the links don't really work doesn't seem like we should bother.

mcollina:
We might want to link the doc file of the SHA we pulled from core.

calvinmetcalf:
covered by  #242

PR #242, opened 2016-11-02
Build 7.0.0

calvinmetcalf:
changes include

- better error messages
- that whole inheritance thing for writable (do we want to remove that?)
- done in _flush accepting a second argument like in next from _transform.

some of these are classified as semver major, though I'd probably only classify them as semver minor as they seem like new features not breaking changes


mcollina:
Seems localtunnel is still not working :(.

mcollina:
I would bump this as a semver-minor, because of the bundling.

calvinmetcalf:
do we want to remove the instanceof stuff that isn't going to help with readable stream and could cause errors ?

mcollina:
I'm +1 on that.

calvinmetcalf:
ok went to add the symbol stuff, but realized that would be fairly brittle to do well, so not sure if we want to do that, though did add some other fixes namley built with 7.1.0, removed the docs,  added a link to the appropriate node docs, remove the duplicate Duplex declare, and made it so tests removed are now actually removed by the build.

calvinmetcalf:
yeah all the node ones have passed and the rest I tested locally

calvinmetcalf:
ok I'm heading out the door soon so I'll publish tomorrow unless I hear any objections

PR #243, opened 2016-11-10
revert bundled dependencies

calvinmetcalf:
looks like there are some issues with globals or something, basically this is breaking yarn yarnpkg/yarn#1774 so I plan on publishing this as 2.2.1 pretty shortly

### DIFF
--- a/package.json
+++ b/package.json
@@ -21,13 +21,6 @@
     "tape": "~4.5.1",
     "zuul": "~3.10.0"
   },
-  "bundledDependencies": [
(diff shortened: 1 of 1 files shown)


mcollina:
Just to clarify, yarn and others __should__ implement bundleDependencies.

calvinmetcalf:
the issue isn't with yarn not handling bundled deps actually, its with *running* yarn, node is looking at the old location so it is likely an upgrading issue

PR #251, opened 2016-12-21
build for 7.3.0

calvinmetcalf:
no changes to streams itself only to tests


mcollina:
yes we have been busy in increasing the coverage of streams internal state machine, so we can catch and avoid regressions. Most of that is @italoacasas, who might want to start tagging along the streams wg ;).

Issue #252, opened 2016-12-23
Next WG meeting - January 2017


mcollina:
How about a Streams WG meeting on the week from the 9th to the 12th?

Agenda:

1. #253, usage of `bind()`, and performance across versions of Node and the browser
2. #249, performance degradation using multiple transforms
3. #254, `Symbol.asyncIterator`

Link to join: https://hangouts.google.com/call/2ko57nnswbf6tlmhcksoc5lrz4e
Link to watch: https://www.youtube.com/watch?v=lthLOvt4QV8

calvinmetcalf:
Works for me

yoshuawuyts:
yup, can do

benjamingr:
I'd like to attend and discuss adding support for `Symbol.asyncIterator` as suggested by @mcollina in https://github.com/nodejs/CTC/issues/53. I'll also open a separate issue.

mcollina:
Here is the doodle: http://doodle.com/poll/ws5xvmrps922a46f.

mcollina:
@benjamingr @mafintosh @calvinmetcalf would you mind filling in the doodle? Today is probably out of the equation by now :D.

benjamingr:
It's 4pm, 5pm and 6pm for me - I'm in GMT + 2. I can make 6pm work by leaving work a little early which is fine.,

mcollina:
All times displayed are CET time. Sigh, I set up the doodle wrong.

Here is the correct new doodle: http://doodle.com/poll/wgbmbcwnyiw7g7nz

benjamingr:
Oh, I thought that the doodle automatically aligns to my timezone. I'll update in the new one.

mcollina:
At this point the two most probable dates are:

* Thursday 12th of January at 19:00 CET
* Friday 13th of January at 18:00 CET

Anyone else that want to partecipate please fill in the doodle.
I'm planning to do Thursday 12th of January, 19:00 CET/ 18:00 GMT, if you all agree.

calvinmetcalf:
sounds good

lrlna:
oh hey i am gonna jump in on this one too; thanks for the ping @mcollina

mcollina:
Confirmed for Thursday 12th of January, 19:00 CET/ 18:00 GMT, as stated in the doodle.
I will post a link.

Can we all compose the agenda, we probably have a few things to talk about.

calvinmetcalf:
[...]

PR #253, opened 2016-12-28
replace bind()

mscdex:
This is anticipation of Writable changes landing in https://github.com/nodejs/node/pull/6533

/cc @calvinmetcalf @mcollina

### DIFF
--- a/build/build.js
+++ b/build/build.js
@@ -42,7 +42,11 @@ function processFile (inputLoc, out, replacements) {
 
     data = data.toString()
     replacements.forEach(function (replacement) {
-      data = data.replace.apply(data, replacement)
+      const regexp = replacement[0]
+      var arg2 = replacement[1]
+      if (typeof arg2 === 'function')
+        arg2 = arg2.bind(data)
+      data = data.replace(regexp, arg2)
     })
     if (inputLoc.slice(-3) === '.js') {
       const transformed = babel.transform(data, {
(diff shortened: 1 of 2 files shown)


calvinmetcalf:
ok this works fine because we put it through babel to deal with the arrow functions, but I wonder why node core couldn't use the arrow function?

mcollina:
`bind()` changed from being a v8 killer to be _faster_ than arrow functions from node v7+.

mscdex:
@mcollina s/arrow functions/inline closures/

mcollina:
@mscdex yes sorry. `bind()` is now faster than using function to wrap `this`.

calvinmetcalf:
so readable stream is intended for use in all versions of node/multiple browsers so I'm not sure the wisdom of including optimizations for older platforms, chrome for instance will be negatively effected by this

mscdex:
@calvinmetcalf Well, it's not just older node versions but also possibly other javascript engines that might not optimize the same things. I personally do not see a straight-forward solution because `readable-stream` is/could be used in so many different environments. If we leave the `bind()` it might work better in some environments worse in others, same goes for reverting its usage. The same also goes for https://github.com/nodejs/readable-stream/pull/255.

I just created these PRs because [@mcollina had mentioned in the upstream node PR](https://github.com/nodejs/node/pull/6533#issuecomment-268741141) that he considered performance changes to be breaking.

mcollina:
@calvinmetcalf using `bind()` in node v0.10/v0.12 and the like can cause a *massive* slowdown. See https://github.com/Level/levelup/pull/139 as an example. If we want to play it safe, we should really benchmark this change on node v0.10-v6, and see if it impacts things. Or, we can do the change and avoid potential breakage (yes, I consider a major slowdown a breakage).

mcollina:
We are going to merge this on the next update, as decided in #252

mcollina:
Part of #262

PR #255, opened 2017-01-03
re-add WriteReq and CorkedRequest constructors

mscdex:
This PR is in anticipation of https://github.com/nodejs/node/pull/10558 (do not land this before the node core PR).

Also, this should land *before* https://github.com/nodejs/readable-stream/pull/253 so that the replacements are done in the correct order.

/cc @mcollina

### DIFF
--- a/build/files.js
+++ b/build/files.js
@@ -183,6 +183,30 @@ const headRegexp = /(^module.exports = \w+;?)/m
     /if \(typeof Symbol === 'function' && Symbol\.hasInstance\) \{/,
     `if (typeof Symbol === 'function' && Symbol.hasInstance && typeof Function.prototype[Symbol.hasInstance] === 'function') {`
   ]
+  , addConstructors = [
+    headRegexp,
+    `$1\n\nfunction WriteReq(chunk, encoding, cb) {
+      this.chunk = chunk;
+      this.encoding = encoding;
+      this.callback = cb;
+      this.next = null;
+    }
+    // It seems a linked list but it is not
+    // there will be only 2 of these for each stream
+    function CorkedRequest(state) {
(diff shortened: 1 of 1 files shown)


calvinmetcalf:
so once  nodejs/node#10558  lands I'll try merging in this and #253

calvinmetcalf:
so I've got a couple questions

1. is this that the perf benefits of object literals vs constructors have changed in more recent versions of v8 or is it that better perfing has shown that object literals were better all along
2. why is this beneficial in node core but NOT in readable stream?

have been out of the loop the last week or 2 so sorry if this has been answered already

mcollina:
> is this that the perf benefits of object literals vs constructors have changed in more recent versions of v8 or is it that better perfing has shown that object literals were better all along

Only most recent versions

> why is this beneficial in node core but NOT in readable stream?

It will cause a slowdown for older node versions.

mcollina:
Part of #262

mcollina:
Merging this have been a mistake. This should be merged when Node.js hits v8.0.0, as the the HTTP changes lands.

calvinmetcalf:
@mcollina there is a merge conflict here, can you rebase it ?

mcollina:
we need this for https://github.com/nodejs/readable-stream/pull/290. I'll see what I can do but I am traveling this week.

calvinmetcalf:
ok I can probably do it too, I'm not 100% on it though like how many libraries are using corked writable streams in the hot path and in older node using readable-streams. Wouldn't they also be having similar if not bigger performance issues from using that older node and they'd get bigger benefits from upgrading ?

mcollina:
Closing as it is now part of #290.

PR #258, opened 2017-02-14
Exclude `stream` module in browsers

RangerMauve:
React Native complains about the call to `require('st' + 'ream')` in `readable.js` with `Unable to resolve module 'stream' from 'bla bla bla/readable-stream/readable.js'`. The `stream` module isn't needed for browsers as far as I can tell, and keeping it as-is means that any module relying on streams within the context of react-native is broken unless you jump through a long line of hoops.

So far I've had issues with this in the context of the `mqtt` and `bl` modules, however there's loads of modules out there that will also have these problems.

### DIFF
--- a/package.json
+++ b/package.json
@@ -39,9 +39,11 @@
     "pipe"
   ],
   "browser": {
-    "util": false,
-    "stream": false
+    "util": false
   },
+  "react-native": {
+    "stream": false 
+  }
   "nyc": {
     "include": [
       "lib/**.js"
--- a/package.json
+++ b/package.json
@@ -39,7 +39,8 @@
     "pipe"
   ],
   "browser": {
-    "util": false
+    "util": false,
+    "stream": false
   },
   "nyc": {
     "include": [
--- a/package.json
+++ b/package.json
@@ -43,7 +43,7 @@
   },
   "react-native": {
     "stream": false 
-  }
+  },
   "nyc": {
     "include": [
       "lib/**.js"


mcollina:
This is very interesting. I do not think we are properly supporting readable-stream in react-native at all. What is the error react-native is complaining about? Is there a cross issue for this on their side?

Can you please delineate the "long line of hoops" that is required without this fix?

@RangerMauve do you have some experience in running unit tests in react-native? I think we should run our suite there as well.

RangerMauve:
> What is the error react-native is complaining about
It's complaining that you're trying to require the `stream` module when it can't find it anywhere in the dependency graph. I guess it somehow figures out that "st" + "ream" is "stream".

>  Is there a cross issue for this on their side
I haven't raised one since I'm not sure what they could do about it other than include node streams in their dependency graph by default, which I doubt they'd want to do.

>  please delineate the "long line of hoops"
I had to manually find all instances of `readable-stream` dependencies in the `node_modules` folder and adding this fix manually to their package.json

> experience in running unit tests in react-native
I'm just starting out with the ecosystem now,, but I think the usual react-native unit tests are going to be different from what's needed here. They recommend using [jest](https://facebook.github.io/jest/) for testing, but I'm not sure where the tests for their tooling lies as their [packager](https://github.com/facebook/react-native/tree/master/packager) doesn't seem to have a test command. They have an [integration tests](https://github.com/facebook/react-native/tree/master/IntegrationTests) folder, but that seems to be used for testing app APIs rather than the packager.

RangerMauve:
Just to clarify, this isn't really a bug in react-native or readable-stream per-se. It's a problem with the fact that react-native doesn't provide node.js built-in modules out of the box, but people still want to use npm modules with it anyway.

RangerMauve:
Another workaround is to `npm install stream` but that [module](https://www.npmjs.com/package/stream) is using a really old version of streams which might cause problem (or not, I'm not sure what the implications are)

mcollina:
@RangerMauve this module underpins all of the streams module on NPM, and it had 40.838.546 downloads **last month** alone. It can be easily used in both browserify and webpack, and possibly all the other packagers for browsers. What you want to change is working successfully everywhere else, so we can't really change it lightly, it might require a major bump for us.

I'm ok in a fix that is specific for react-native, but changing the logic for everyone is a 👎  for me. However, I would really like to see this available within react-native.

RangerMauve:
Did some more investigating and it seems that in addition to the `browser` field, they also support [a `react-native` field](https://github.com/facebook/react-native/blob/master/packager/src/node-haste/Package.js#L141)

If I modify the PR to use that field instead of the `browser` field, it wouldn't affect anything outside of react-native. Would that de-escalate this down to a patch?

As for including streams in react-native, I would really love that to happen. But from what I can tell they seem to be more interested in emulating browsers than Node.js, so it's more likely that they'd eventually introduce WHATWG streams or even observables or async generators instead of node streams

mcollina:
When I mean node streams, I mean all the modules that are there on NPM. I don't really care what's underneath, we can add wrapper under whatever is available in there.

RangerMauve:
Yeah, I just tested this out and it solves the issue without affecting any existing browserify or webpack builds.

package.json
```diff
@@ -41,6 +41,9 @@
   "browser": {
     "util": false
   },
+  "react-native": {
+    "stream": false 
+  }
```
mcollina: there is a `,` missing after this

RangerMauve: Whoops, that's what I get for using the github editor. Pushed a fix

mcollina:
@calvinmetcalf what do you think of the latest fix?

calvinmetcalf:
have you tested this on react native ?

RangerMauve:
I'm actively using this in a react native project right now. I manually edited the modules I had installed to have this fix.

calvinmetcalf:
I can publish tommorow

RangerMauve:
Woot, thank you!

calvinmetcalf:
done

parshap:
@RangerMauve: I'm curious how you got this working -- when I tried this, it causes `require('st' + 'ream')` to return an empty object `{}` which then causes an exception to be thrown [when calling `inherits()`](https://github.com/nodejs/readable-stream/blob/4dbe6e20051e44025d3cbec3ed9603483d8b27cb/lib/_stream_readable.js#L61).

RangerMauve:
You're right! I thought it was working because I was able to compile my app, but apparently the library using `readable-stream` wasn't actually instantiating anything, and thus I wasn't seeing any runtime errors. Once I fixed that issue it started throwing at the `inherits()` line. Right now I've been de-focused from react-native for a bit and will have to get back to it in a week or two.

PR #264, opened 2017-03-14
Fully revert #255 as not released yet in core.

mcollina:
cc @calvinmetcalf.

This revert #255 completely.


mcollina:
@calvinmetcalf if you are ok, I'm going to release anyway. It seems it's impossible right now to get localtunnel to work.

calvinmetcalf:
👍 go for it

PR #266, opened 2017-03-17
Fixed 'Invalid non-string/buffer chunk' error when using Uint8Array

eavidan-0:
Using 'new Buffer' on a Uint8Array does not create a Buffer object, but adding a Uint8Array chunk is supported by the library, so I added a check to the validation to allow this

### DIFF
--- a/lib/_stream_readable.js
+++ b/lib/_stream_readable.js
@@ -394,11 +394,10 @@ Readable.prototype.read = function (n) {
 };
 
 function chunkInvalid(state, chunk) {
-  var er = null;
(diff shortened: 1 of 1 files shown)


mcollina:
We pull our changes from Node.js core, so we cannot accept this one here.
You should send a PR to https://github.com/nodejs/node.

How does this relates to https://github.com/nodejs/node/pull/11608?

cc @addaleax

addaleax:
> How does this relates to nodejs/node#11608?

They try to achieve the same thing (I think), but this approach would just let `Uint8Array`s pass through to the `_read()` method instead of adjusting its prototype. I don’t think that’s a good idea.

Issue #260, opened 2017-02-28
require('stre'+'am') doesn't do what I thought it did


calvinmetcalf:
instead of only including stream if somebody else included it it apparently just doesn't include it in browserify vs node.  This tells us a couple things

1. nobody is doing readable-stream instance of require('stream') in the browser
2. we could theoretically just replace this with a "browser":{"stream":false}
3. webpack is doing what I thought browserify did, but they are doing it wrong

calvinmetcalf:
see webpack/webpack#4378

mcollina:
Thanks @calvinmetcalf for taking this heads-on.
I'm 👍  on `"browser": { "stream": false }`.

Should we discuss this on the next wg meeting and release this ASAP?

calvinmetcalf:
yeah

parshap:
I've struggled with the `require('st'+'ream')` calls while trying to get readable-stream to run in React Native. Here's what I've observed:

1. **When using Browserify, `require('st'+'ream')` throws an exception.** This is because Browserify *does not* resolve `'st' + 'ream'` at compile-time, so the module is not available at run-time. This is true even when another module has required `stream` before because the `require` function in Browserify is local to the module and only gives access to the dependencies resolved at compile-time.

1. **When using React Native Packager, `require('st'+'ream')` returns an empty object `{}`.** This is because React Native Packager *does* resolve `'st' + 'ream'` at compile-time and [`stream: false` is set in `package.json`][stream-false], which [according to the `package.json` `browser` field spec][browser spec], should cause the module to resolve to an empty object `{}`.

This results in readable-stream throwing an exception and crashing the app. My current solution is to [patch readable-stream to remove `require('st' + 'ream')` calls][patch] in a `postinstall` hook.

Going forward, my suggestion is to:

1. Break `instanceof` compatibility with Node's builtin `Stream`. It wasn't working in the first place and no one complained.

1. If breaking this compatibility is not desired, set `stream: false` for both the `browser` and `react-native` fields in `package.json`. Call `require('stream')` and check that it returns `Stream` and not the empty object before using it.

I can submit a PR for this. 

[stream-false]: https://github.com/nodejs/readable-stream/blob/4dbe6e20051e44025d3cbec3ed9603483d8b27cb/package.json
[browser spec]: https://github.com/defunctzombie/package-browser-field-spec/tree/4f296871cee64e60124841c06c06511885152f19#ignore-a-module
[...]

PR #267, opened 2017-03-17
remove bizzare circular deps #260

calvinmetcalf:
remember this commit a477a21474c0db82e34ee0bb1c0ff3b05dae1e41 ? yeah turns out it's a breaking change, but this didn't actually break anything in those 2 years, plus there has been a breaking change since then we can probably take it out.

### DIFF
--- a/package.json
+++ b/package.json
@@ -39,7 +39,8 @@
     "pipe"
   ],
   "browser": {
-    "util": false
+    "util": false,
+    "./readable.js": "./readlable-browser.js"
   },
   "react-native": {
     "stream": false
--- a/package.json
+++ b/package.json
@@ -40,7 +40,7 @@
   ],
   "browser": {
     "util": false,
-    "./readable.js": "./readlable-browser.js"
+    "./readable.js": "./readable-browser.js"
   },
   "react-native": {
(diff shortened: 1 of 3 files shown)


readable.js
```diff
@@ -1,16 +1,14 @@
...
 exports.PassThrough = require('./lib/_stream_passthrough.js');
 
-if (!process.browser && process.env.READABLE_STREAM === 'disable' && Stream) {
-  module.exports = Stream;
+if (!process.browser) {
+  var Stream = require('st' + 'ream');
+  if (process.env.READABLE_STREAM === 'disable' && Stream) {
+    module.exports =  Stream;
+  }
 }
```
calvinmetcalf: yes

package.json
```diff
@@ -39,7 +39,8 @@
     "pipe"
   ],
   "browser": {
-    "util": false
+    "util": false,
+    "./readable.js": "./readlable-browser.js"
```
mcollina: this is probably a typo.

calvinmetcalf: trying to do too many things at once thanks

calvinmetcalf:
closing in favor of #269

PR #268, opened 2017-03-28
Avoid caret during installation, and added node v0.6 to .travis.yml

mcollina:
For some context, I would like mysql to use the latest readable-stream: https://github.com/mysqljs/mysql/pull/1677

### DIFF
--- a/package.json
+++ b/package.json
@@ -4,7 +4,7 @@
   "description": "Streams3, a user-land copy of the stream library from Node.js",
   "main": "readable.js",
   "dependencies": {
-    "buffer-shims": "^1.0.0",
+    "buffer-shims": "~1.0.0",
     "core-util-is": "~1.0.0",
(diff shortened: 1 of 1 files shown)


calvinmetcalf:
dependencies of dependencies are also an issue, which is why I did it this way

mcollina:
Updated, this should work also on older versions of NPM.

mcollina:
@calvinmetcalf yes. But it might allow the mysql module to upgrade to v2. We don't expect buffer-shims to change at any point in time either.

AdriVanHoudt:
I don't see the node 0.6 travis change in the files changed as stated in the title or is that just me?

mcollina:
I updated the title. We cannot really run our tests on Node v0.6.
Depending using `~` simplify backporting to old NPM versions. The tests passes, but our own test dependencies cannot be installed there anymore.

PR #269, opened 2017-03-28
Maintain the same interface when using READABLE_STREAM=disable

mcollina:
This would need to be backported to the 1.0 and 1.1 lines, as they suffer from the same bug.

Fixes https://github.com/nodejs/node/issues/12006

### DIFF
--- a/transform.js
+++ b/transform.js
@@ -1,13 +1,1 @@
-var Transform = require("./lib/_stream_transform.js")
-
-if (!process.browser && process.env.READABLE_STREAM === 'disable') {
-  var Stream = (function (){
-    try {
-      return require('st' + 'ream'); // hack to fix a circular dependency issue when used with browserify
-    } catch(_){}
-  }());
-
-  module.exports = Stream && Stream.Transform || Transform
-}
-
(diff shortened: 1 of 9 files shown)


calvinmetcalf:
you probably want to build this on top of #260

mcollina:
@calvinmetcalf updated and refactored.

I'm wondering if we should change this replacement as well for browsers:
https://github.com/nodejs/readable-stream/blob/master/lib/_stream_readable.js#L27-L36.

and flag `"stream": false`  in the  `"browser"`  field in `package.json`?

calvinmetcalf:
yes good call

transform.js
```diff
@@ -1 +1 @@
-module.exports = require("./lib/_stream_transform.js")
+module.exports = require('./readable').Transform
```
calvinmetcalf: could this theoretically pull in more then it did previously? probably just passthrough right

mcollina:
@nodejs/streams may I get some LGTMs so we can release this?

I think this is a patch release, as we are fixing a bunch of things.

Issue #277, opened 2017-04-07
regression v2.2.7 and v2.2.8 are broken on some webpack / browserify deployments


mcollina:
Original issue: https://github.com/mqttjs/MQTT.js/issues/596.
It was fixed by blocking the version of readable-stream to 2.2.6.

cc @calvinmetcalf

mcollina:
I cannot reproduce myself really, so something else should be going on.

I will leave it open, in case @CVarisco would like to produce a full example that is reproducing the problem.

I think the problem is that in https://github.com/nodejs/readable-stream/blob/master/lib/_stream_readable.js#L33, in some cases `Stream` is `{}` instead of null.

RangerMauve:
@mcollina If the problem is Stream being an empty object, that's definitely related to the PR I sent in #258  and affect people trying to use React-Native. In https://github.com/mqttjs/MQTT.js/issues/573 I proposed to use the following change to see if it'd fix things:

var Stream;
(function () {
  try {
    Stream = require('st' + 'ream');
  } catch (_) {} finally {
    if (!Stream) Stream = require('events').EventEmitter;
  }
  // This is the addiition
  if(!Stream.on) Stream =  require('events').EventEmitter;
})();

I haven't had time to test this out, however.

calvinmetcalf:
@RangerMauve in react native, is process.browser true?

RangerMauve:
@calvinmetcalf No, it's the packager which causes it to happen by looking in the [react-native property in package.json](https://github.com/nodejs/readable-stream/blob/master/package.json#L47). It has no effect on bundlers which don't use the `react-native` field.

I don't think react-native even defines a process global, actually.

calvinmetcalf:
so I just opened #278 which uses process.browser to tell if it should import stream or go with event emitter, I'm not sure of the behavior of react-native but I'm fairly confident that if it already worked, it has to have some form of process global just not sure what it is.

RangerMauve:
In order to get it to run in React-Native I had to define `process` and `Buffer` globals, for process I was using  [this module](https://www.npmjs.com/package/process), and React-Native was likely loading the browser version of it since it respects the `browser` field in package.json

calvinmetcalf:
ok so then my pull will likely work there

mcollina:
another duplicate https://github.com/maxogden/websocket-stream/issues/114.
Il giorno ven 7 apr 2017 alle 18:37 Calvin Metcalf <notifications@github.com>
[...]

PR #278, opened 2017-04-07
Simpler event stream sub

calvinmetcalf:
fix for #277, hopfully


mcollina:
LGTM and let's see what happens.

calvinmetcalf:
wait actually I have a better idea

calvinmetcalf:
we already are successfully using the package.browser switch, so lets use it for this

calvinmetcalf:
ok @mcollina check this out

RangerMauve:
@calvinmetcalf You'll want to apply these changes to the `react-native` field, too, or get rid of it to use the browser version instead.

calvinmetcalf:
ok good call

RangerMauve:
Err, wait, it'll still bark on loading the `stream` module. Maybe copy the whole browser object, and for React-Native add `"stream": false`. Since it needs all the changes for browser compatibility, plus it needs the stream module to be explicitly excluded, else the `require("st" + "ream")` call will crash the bundler.

calvinmetcalf:
there shouldn't be any stream calls anymore in the browser code

RangerMauve:
You're right, sorry about that. :P

calvinmetcalf:
so it's good ?

calvinmetcalf:
@mcollina the idea isn't that we need the same api, it's that `instanceof stream` checks should succeed

mcollina:
LGTM

PR #279, opened 2017-04-16
README: link to specific api docs about streams

Haroenv:


calvinmetcalf:
do we need to change this in the build scripts @mcollina ?

mcollina:
The build script just applies a regexp, so no, we can merge this.

Issue #281, opened 2017-04-30
Exclude `buffer-shims` from browser build


feross:
The browser implementation of [`buffer`](https://github.com/feross/buffer) (used by `browserify` and `webpack`) supports the new buffer methods (`Buffer.from`, etc.) so there is no need to use `buffer-shims` in the browser.

Let's exclude `buffer-shims` and just use `Buffer.from`, etc. directly in the browser to save sending 7.44 kB of extra code that is never used.

calvinmetcalf:
I was actually thinking of switching to safe buffer

mcollina:
@feross we need `buffer-shims` (or `safe-buffer`) because we must support old Node.js version, not just the browser. How should we do that in browserify?

RangerMauve:
Maybe you could have a file called `buffer-impl.js` which in the browser would return `global.Buffer` and in Node return `safe-buffer`? Or better yet, have a browser alias for `safe-buffer` which points to a file which exports `global.Buffer`

calvinmetcalf:
I think [safe-buffer](https://github.com/feross/safe-buffer) does everything we need, I probably wouldn't have written buffer-shim if I know about it

feross:
> @mcollina we need buffer-shims (or safe-buffer) because we must support old Node.js version, not just the browser. How should we do that in browserify?

I think you probably already figured it out, but just documenting for others. In the browser `require('safe-buffer')` is just an alias for `require('buffer')` which already has all the needed methods. See [`browser.js`](https://github.com/feross/safe-buffer/blob/6ed80179e4847e49781de0587eedae5d7d658949/browser.js).

PR #282, opened 2017-05-01
use safe-buffer

calvinmetcalf:
see #281


mcollina:
We need to update https://github.com/rvagg/string_decoder as well.

calvinmetcalf:
rvagg/string_decoder#3

PR #286, opened 2017-05-28
Updated to node v7.10.0.

mcollina:
cc @nodejs/streams

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -2,3 +2,4 @@ node_modules/
 .zuul.yml
 .nyc_output
 coverage
+package-lock.json
--- a/.travis.yml
+++ b/.travis.yml
@@ -44,6 +44,8 @@ matrix:
     env: TASK=test
   - node_js: 7
     env: TASK=test
+  - node_js: 8
+    env: TASK=test
   - node_js: 6
     env: TASK=browser BROWSER_NAME=ie BROWSER_VERSION="9..latest"
   - node_js: 6
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,26 +2,40 @@ sudo: false
 language: node_js
 before_install:
   - npm install -g npm@2
-  - npm install -g npm@4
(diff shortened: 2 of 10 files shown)


mcollina:
@calvinmetcalf there was a verify high number of breakages due to npm and some things that were not transpiled.

mcollina:
@calvinmetcalf CI is badly broken :( https://travis-ci.org/nodejs/readable-stream/jobs/237500260

calvinmetcalf:
ah looks like they stoped supporting 0.10?

mcollina:
@zkat can you have a look here? it used to work :(.

zkat:
@mcollina we do not support, at all, any versions of node older than node@4. Only node versions still in maintenance are targeted, and we've started intentionally breaking compatibility now.

mcollina:
@zkat You mean that `npm@lts` is not supported anymore? what about `npm@latest-3`?

zkat:
@mcollina I guess lts and 3 would probably still work, but this stuff isn't really supposed to run on `node@0.10`. It looks like we officially took out `0.10` support with `npm@4.0`, and dropped `0.12` support as of `npm@4.1.2`.

mcollina:
@zkat we are now using npm@2 and npm@latest-3 in the unsupported versions of Node. Things are passing again now :).

mcollina:
@calvinmetcalf I disabled the new `test/parallel/test-stream-unpipe-event.js` in Node v0.8 as it was not passing. I don't think there is any value in fixing it.

calvinmetcalf:
ok what was the issue with unpipe events ?

mcollina:
This is the relative PR: https://github.com/nodejs/node/pull/11876. I do not know why it does not pass on node 0.8.

calvinmetcalf:
ok looks like the test may have been wonky? I don't know lgtm

PR #288, opened 2017-06-06
Use tilde with safe buffer

mcollina:
cc @dougwilson


mcollina:
Node.js is passing, merging and releasing.

PR #290, opened 2017-06-09
Node 8

calvinmetcalf:


calvinmetcalf:
ok I need to fix the Stream._isUint8Array and similar stuff

mcollina:
There is a PR waiting to me merged for this.

cc @addaleax for the other things

calvinmetcalf:
I know how to add them I just didn't get around to it before I went home

calvinmetcalf:
and we are green (though I need to rebase it on top of the other pull)

mcollina:
@calvinmetcalf I've added #255 commit in here. It should be good to go.

calvinmetcalf:
ok bumped it so it should hopfully build

mcollina:
I think we should release it as 2.3.0. Let's do this next week, so that @nodejs/streams might have time to review it (I would also be at home and not traveling, in case we need to last-minute bugfixes)

calvinmetcalf:
ok sounds good

addaleax:
I looked over the patch and nothing seemed to stick out … I might not have time for a full review but this seems fine to me.

addaleax:
We might want to include something like https://github.com/nodejs/node/pull/13760.

mcollina:
I think we should wait for that to be released

calvinmetcalf:
ok I'll plan on publishing this soon

PR #295, opened 2017-06-21
Import from Node 8.1.2.

### LINKED ISSUE #293: 2.3.0 breaks PhantomJS 1.x, IE 7 & 8
Mocha is using Browserify to bundle the project's tests for the browser. Just the other day we started getting errors [due to use of `Function.prototype.bind` in PhantomJS 1](https://travis-ci.org/mochajs/mocha/jobs/243848538#L327) and due to use of [`Object.keys` (in the `safe-buffer` dependency) in Internet Explorer 7 and 8](https://travis-ci.org/mochajs/mocha/jobs/243848538#L632). (I strongly suspect that `Function.prototype.bind` would be an issue in IE 7 &/ 8 as well except that they fail on the `Object.keys` issue first.) [Specifically using `readable-stream@2.2.11` fixes both issues](https://github.com/mochajs/mocha/pull/2898).

(Between the fact that we can work around this as long as nothing is forcing 2.3 into the browser bundle and the fact that we're testing unusually old browsers, I'd understand if this is considered a non-issue; but I wanted to bring it to your attention for what it's worth.)

mcollina:
Fixes #293 (our `bind()` part)

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,6 +1,6 @@
 # readable-stream
 
-***Node-core v8.1.0 streams for userland*** [![Build Status](https://travis-ci.org/nodejs/readable-stream.svg?branch=master)](https://travis-ci.org/nodejs/readable-stream)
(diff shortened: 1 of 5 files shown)


calvinmetcalf:
this looks good, the error is unrelated

PR #298, opened 2017-06-21
Forward-port of nodejs#13850

mcollina:
So, here is the forward-port of that PR. It will likely be part of Node 8.1.3, which will likely be released next week. I'm not sure if it's worth porting it forward here.

https://github.com/nodejs/node/pull/13850

cc @phated

### DIFF
--- a/lib/_stream_writable.js
+++ b/lib/_stream_writable.js
@@ -408,15 +408,26 @@ function doWrite(stream, state, writev, len, chunk, encoding, cb) {
 
 function onwriteError(stream, state, sync, er, cb) {
   --state.pendingcb;
-  if (sync) processNextTick(afterError, stream, state, cb, er);else afterError(stream, state, cb, er);
 
-  stream._writableState.errorEmitted = true;
-  stream.emit('error', er);
-}
-
-function afterError(stream, state, cb, err) {
-  cb(err);
-  finishMaybe(stream, state);
+  if (sync) {
+    // defer the callback if we are being called synchronously
+    // to avoid piling up things on the stack
+    processNextTick(cb, er);
+    // this can emit finish, and it will always happen
+    // after error
+    processNextTick(finishMaybe, stream, state);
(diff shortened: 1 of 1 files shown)


calvinmetcalf:
nice but we have to be carefull because any other pull will likely revert it

lib/_stream_writable.js
```diff
@@ -408,15 +408,26 @@ function doWrite(stream, state, writev, len, chunk, encoding, cb) {
...
-function afterError(stream, state, cb, err) {
-  cb(err);
-  finishMaybe(stream, state);
+  if (sync) {
+    // defer the callback if we are being called synchronously
+    // to avoid piling up things on the stack
+    process.nextTick(cb, er);
+    // this can emit finish, and it will always happen
+    // after error
+    process.nextTick(finishMaybe, stream, state);
```
calvinmetcalf: this needs to be processNextTick

lib/_stream_writable.js
```diff
@@ -408,15 +408,26 @@ function doWrite(stream, state, writev, len, chunk, encoding, cb) {
...
-  stream.emit('error', er);
-}
-
-function afterError(stream, state, cb, err) {
-  cb(err);
-  finishMaybe(stream, state);
+  if (sync) {
+    // defer the callback if we are being called synchronously
+    // to avoid piling up things on the stack
+    process.nextTick(cb, er);
```
calvinmetcalf: needs to be processNextTick

mafintosh:
lgtm, other than the other comments, with the caveats that i'm 100% up to date with the new mechanics. 

Are there any guarantees now about close firing before after finish etc?

mcollina:
> Are there any guarantees now about close firing before after finish etc?

I have no idea. I think the two things are loosely correlated in duplexes. Given the rate of unexpected breakage, I think we have poor coverage of the state machine and the ecosystem.
However, let's tackle one regression at a time :/.

Issue #302, opened 2017-06-27
UInt8Array check slowing things down


mcollina:
@addaleax the supported version of UInt8Array is slowing things down significantly here.
Can we implement the check in a faster way? 

cc https://bugs.chromium.org/p/v8/issues/detail?id=6519#c2

PR #304, opened 2017-06-27
Partially solves perf issues with the UInt8Array checks.

mcollina:
Fixes https://github.com/nodejs/readable-stream/issues/302

This is a partial fix.

Node 6.11, using readable-stream 2.2.11:

```
benchThrough2*10000: 5148.149ms
```

Node 6.11, using readable-stream 2.3.2:
```
benchThrough2*10000: 18112.640ms
```

Node 6.11, using this patch:

```
benchThrough2*10000: 6491.796ms
```

I think we can do better.

cc @bmeurer

### DIFF
--- a/build/files.js
+++ b/build/files.js
@@ -185,7 +185,7 @@ function _uint8ArrayToBuffer(chunk) {
    return Buffer.from(chunk);
 }
 function _isUint8Array(obj) {
-  return Buffer.isBuffer(obj) || (typeof obj !== 'string' && obj instanceof OurUint8Array);
+  return Buffer.isBuffer(obj) || obj instanceof OurUint8Array;
 }
 /*</replacement>*/
   `
--- a/build/files.js
+++ b/build/files.js
@@ -184,7 +184,7 @@ function _uint8ArrayToBuffer(chunk) {
    return Buffer.from(chunk);
 }
 function _isUint8Array(obj) {
-  return Object.prototype.toString.call(obj) === '[object Uint8Array]' || Buffer.isBuffer(obj);
+  return Buffer.isBuffer(obj) || (typeof obj !== 'string' && Object.prototype.toString.call(obj) === '[object Uint8Array]');
 }
 /*</replacement>*/
   `
--- a/build/files.js
+++ b/build/files.js
@@ -180,11 +180,12 @@ const headRegexp = /(^module.exports = \w+;?)/m
       /(?:var|const) Buffer = require\('buffer'\)\.Buffer;/
     , `/*<replacement>*/
   const Buffer = require('safe-buffer').Buffer
+  const OurUint8Array = global.Uint8Array || function () {}
 function _uint8ArrayToBuffer(chunk) {
    return Buffer.from(chunk);
 }
 function _isUint8Array(obj) {
-  return Buffer.isBuffer(obj) || (typeof obj !== 'string' && Object.prototype.toString.call(obj) === '[object Uint8Array]');
+  return Buffer.isBuffer(obj) || (typeof obj !== 'string' && obj instanceof OurUint8Array);
 }
 /*</replacement>*/
   `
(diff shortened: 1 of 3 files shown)


mcollina:
Using `instanceof` results in a better:

```
benchThrough2*10000: 5649.387ms
```

build/files.js
```diff
@@ -180,11 +180,12 @@ const headRegexp = /(^module.exports = \w+;?)/m
       /(?:var|const) Buffer = require\('buffer'\)\.Buffer;/
     , `/*<replacement>*/
   const Buffer = require('safe-buffer').Buffer
+  const OurUint8Array = global.Uint8Array || function () {}
 function _uint8ArrayToBuffer(chunk) {
    return Buffer.from(chunk);
 }
 function _isUint8Array(obj) {
-  return Object.prototype.toString.call(obj) === '[object Uint8Array]' || Buffer.isBuffer(obj);
+  return Buffer.isBuffer(obj) || (typeof obj !== 'string' && obj instanceof OurUint8Array);
```
bmeurer: Why not `typeof obj === 'object'`?

lib/_stream_readable.js
```diff
@@ -54,11 +54,12 @@ var Stream = require('./internal/streams/stream');
 // properly optimized away early in Ignition+TurboFan.
 /*<replacement>*/
 var Buffer = require('safe-buffer').Buffer;
+var OurUint8Array = global.Uint8Array || function () {};
 function _uint8ArrayToBuffer(chunk) {
   return Buffer.from(chunk);
 }
 function _isUint8Array(obj) {
-  return Object.prototype.toString.call(obj) === '[object Uint8Array]' || Buffer.isBuffer(obj);
+  return Buffer.isBuffer(obj) || typeof obj !== 'string' && obj instanceof OurUint8Array;
```
bmeurer: Same.

lib/_stream_writable.js
```diff
@@ -80,11 +80,12 @@ var Stream = require('./internal/streams/stream');
 
 /*<replacement>*/
 var Buffer = require('safe-buffer').Buffer;
+var OurUint8Array = global.Uint8Array || function () {};
 function _uint8ArrayToBuffer(chunk) {
   return Buffer.from(chunk);
 }
 function _isUint8Array(obj) {
-  return Object.prototype.toString.call(obj) === '[object Uint8Array]' || Buffer.isBuffer(obj);
+  return Buffer.isBuffer(obj) || typeof obj !== 'string' && obj instanceof OurUint8Array;
```
bmeurer: Same.

build/files.js
bmeurer: Actually now that I think about it, the `typeof` check is just unnecessary complexity. `instanceof` itself already does a type check anyways (as one of the first steps).

build/files.js
```diff
@@ -180,11 +180,12 @@ const headRegexp = /(^module.exports = \w+;?)/m
       /(?:var|const) Buffer = require\('buffer'\)\.Buffer;/
     , `/*<replacement>*/
   const Buffer = require('safe-buffer').Buffer
+  const OurUint8Array = global.Uint8Array || function () {}
```
bmeurer: Another thing: Use `var` instead of `const` here (and in the other files below), to ensure that the `instanceof` optimization definitely triggers.

mcollina: we are passing through babel anyway, these are all vars.

mcollina:
@bmeurer thanks, updated.

kgryte:
The move to using `instanceof` will fail for cross-realm `Uint8Array` objects (e.g., from iframes or, depending on the setup/Node.js version, different node V8 vm's). Is that a concern?

In contrast, the `Object.prototype.toString()` approach supports cross-realm objects.

mcollina:
I do not think so. Considering the cost of using the `toString()` approach, I think it's a trade-off that we must accept.

bmeurer:
Ideally there'd be an *instance type* check for `Uint8Array` and friends. `instanceof` doesn't really check whether the object is an `Uint8Array`, but rather if it has a certain prototype, which is obviously realm specific.

PR #307, opened 2017-07-04
Add nodejs supported version in package

allevo:
The `.travis.yml` describes the nodejs supported version.
This is not reflected in `package.json`.

(https://docs.npmjs.com/files/package.json#engines)

### DIFF
--- a/package.json
+++ b/package.json
@@ -38,6 +38,9 @@
     "stream",
     "pipe"
   ],
+  "engines": {
+    "node": ">=0.8"
+  },
   "browser": {
     "util": false,
     "./readable.js": "./readable-browser.js",


mcollina:
I'm 👎  on this change. Why is it needed?

allevo:
- The first things I see in a npm module are the supported nodejs version in order to define the supported version in the my package.
- Using `npm` the only way to define them is in `package.json`. And the test are run defining the nodejs versions (in `.travis.yml`). There's no reason to don't explicitly declare the supported nodejs version. 
- The main aim of this package is to achieve the back-compatibility through the stream versions.

That's why I thought it is useful.

calvinmetcalf:
@allevo one of the issues we face in older node is that some of the testing infrastructure doesn't work in older versions of node that readable-stream might, so doing this may break older versions of node that actually do work even if the tests don't

allevo:
I use `0.8` version because I found it in `.travis.yml` file. `tap` works only with `0.8` or higher. This don't imply that the older version don't works. I pulled random the number.

But the other considerations are still valid.

Is there a known nodejs version that works with `redeable-stream`?

mcollina:
Closing for now. Feel free to send this to #344, we can ship it there.

PR #314, opened 2017-10-04
fix _stream_writeable to not call functions on undefined properties

### LINKED ISSUE #313: process.version.slice - undefined is not an object
It appears that not all environments that are using this library pass a `process.version`.

Some of the libraries that I use in React Native are using this library at run-time and are not passing a version.  Since version is undefined, the slice method fails.

gwmccull:
This fixes #313

### DIFF
--- a/lib/_stream_writable.js
+++ b/lib/_stream_writable.js
@@ -54,6 +54,9 @@ function CorkedRequest(state) {
 /* </replacement> */
 
 /*<replacement>*/
+if (!process.version) {
+	process.version = '';
+}
 var asyncWrite = !process.browser && ['v0.10', 'v0.9.'].indexOf(process.version.slice(0, 5)) > -1 ? setImmediate : processNextTick;
 /*</replacement>*/
 


mcollina:
Can you update the build scripts (https://github.com/nodejs/readable-stream/tree/master/build) as we pull the code from node core releases?

gwmccull:
Hi @mcollina,

I'd be happy to make changes but I'm not sure which build scripts you're asking about.

Can you point me towards the one that needs to be updated? Or, is one of them failing that I could run and see the output?


Thanks,

Garrett

mcollina:
Here it is: https://github.com/nodejs/readable-stream/blob/d6c391d7101946c8b8b97914fc46fd3322c450d1/build/files.js#L147-L149

gwmccull:
I found out that you can polyfill `process.version` by creating a globals file and specifying the missing `process` properties there. That's probably the best way to go rather than try to "fix" all of the Node libraries

I'm going to close this PR for now.  Feel free to re-open if you think it's really necessary.

quocson1:
image
how to fix

PR #317, opened 2017-12-11
Update process-nextick-args to most recent release. Fixes #316

### LINKED ISSUE #316: sinon/lolex fake timers broken when used with this
Consider the following mocha test file

```js
var lolex = require('lolex');
var through = require('through2');

describe('a bug', function() {
  var clock;
  before(function() {
    clock = lolex.install({toFake: [ 'setImmediate', 'nextTick' ]});
  });
  after(function() {
    clock.uninstall();
  });

  it('does not work well with lolex', function(done) {
    var stream = through.obj();
    stream.on('data', function() {
      stream.on('end', function() {

        console.log('creating stream2')
        var stream2 = through.obj();
        stream2.on('data', function() {
          stream2.on('end', function() {
            console.log('OK')
            done();            
          });
          setImmediate(function() {
            stream2.end()
          });
        });
        stream2.emit('data')

      });
      stream.end();
    });
    stream.emit('data');

    clock.runAll()
  });

});

```

When the `clock` calls are removed and this runs using regular time this runs and finishes as expected and without errors. However when running with lolex `install`ed it just hangs after logging `creating stream2`. This is due to https://github.com/nodejs/readable-stream/blob/d6c391d7101946c8b8b97914fc46fd3322c450d1/lib/_stream_readable.js#L983 which retains a reference to the non swapped out version of `process.nextTick` due to use of `process-nextick-args` https://github.com/nodejs/readable-stream/blob/d6c391d7101946c8b8b97914fc46fd3322c450d1/lib/_stream_readable.js#L26

A simple solution can be swapping `require('process-nextick-args')` with `require('./process-nextick-args')` which can look something like this:

```js
//  ./process-nextick-args
var nextTickNeedsPatching = (
  !process.version ||
  process.version.indexOf('v0.') === 0 ||
  process.version.indexOf('v1.') === 0 && process.version.indexOf('v1.8.') !== 0
);

var processNextTick = function(fn, a, b, c) {
  return !nextTickNeedsPatching ? process.nextTick(fn, a, b, c) :
  process.nextTick(function() {
    fn(a, b, c);
  });
  
}
```

Does that sound like a reasonable solution?

kolodny:


mcollina:
You will need to update the scripts in `build`, as all this code is auto-generated. I plan to do it as soon as possible anyway.

kolodny:
Sounds good, should I close this PR and you'll start a new one or did you plan on adding commits to this one?

mcollina:
I’d suggest to do a new one, just build the same node version that is in the README.

kolodny:
OK, I'll let you handle this one 😉

PR #320, opened 2018-02-08
Build 8.9.4

### LINKED ISSUE #316: sinon/lolex fake timers broken when used with this

mcollina:
As titled.

I have also updated process-nextick-args.
Fixes #316.


mcollina:
```
> [secure]@2.3.3 write-zuul /home/travis/build/nodejs/[secure]
> printf "ui: tape
browsers:
  - name: $BROWSER_NAME
    version: $BROWSER_VERSION
">.zuul.yml
- testing: microsoftedge @ Windows 10: 16
- restarting: <microsoftedge 16 on Windows 10>
- restarting: <microsoftedge 16 on Windows 10>
- failed: <microsoftedge 16 on Windows 10> (0 failed, 0 passed)
1 browser(s) failed
npm ERR! code ELIFECYCLE
npm ERR! errno 1
npm ERR! [secure]@2.3.3 browser: `npm run write-zuul && zuul --browser-retries 2 -- test/browser.js`
npm ERR! Exit status 1
npm ERR! 
npm ERR! Failed at the [secure]@2.3.3 browser script.
npm ERR! This is probably not a problem with npm. There is likely additional logging output above.
npm ERR! A complete log of this run can be found in:
npm ERR!     /home/travis/.npm/_logs/2018-02-08T09_57_28_455Z-debug.log
```

localtunnel/saucelabs is not really working atm.

vweevers:
@mcollina FWIW memdown had similar issues, that disappeared when we switched to Sauce Connect (https://github.com/Level/memdown/pull/81) - though it depends on an as-of-yet unmerged change to `zuul`.

mcollina:
@vweevers considering that zuul is not really well maintained, how about we spin up a new project for that? I would prefer to depend on something from npm and not from a github account.

vweevers:
SGTM! Can we reduce its scope (Sauce Connect only) to lower the maintenance cost?

mcollina:
> SGTM! Can we reduce its scope (Sauce Connect only) to lower the maintenance cost?

Yes definitely.

yoshuawuyts:
Scaling CI down to just Sauce Connect sounds good! ~ have been frustrated with Zuul for a while too now; having a more reliable solution is definitely a step forward.

vweevers:
I should have knocked on wood.. Sauce Labs is now also failing for memdown, with an error similar to https://github.com/nodejs/readable-stream/pull/320#issuecomment-364096828. But I think forking zuul to work with Sauce Connect is still a good idea, as it at least removes localtunnel as a point of failure. If you agree I'll open an repo to continue discussion there.

mcollina:
@vweevers I also spoke with @feross whom agreed on helping out. I would say go ahead and create a repo and add all of us, I think we can build something that it's easy to maintain.

Issue #327, opened 2018-03-27
Events module not in package.json


mahatosourav91:
Hi, I am getting the following error,
error: bundling failed: Error: Unable to resolve module `events` from ``/Users/node_modules/redux-persist-transform-encrypt/node_modules/readable-stream/lib/_stream_readable.js`: Module does not exist in the module map`

I believe events module is missing in **package.json**

mcollina:
The events module is provided by Node.js or your bundler. Which bundler are you using? You might need to configure it accordingly.

PR #328, opened 2018-03-27
events added in package.json

mahatosourav91:
events added in package.json
As per the issue [327](https://github.com/nodejs/readable-stream/issues/327)


mcollina:
This should not be needed, and it should be handled by your bundler.

PR #331, opened 2018-04-02
Don't add the Array.prototype.forEach replacement to files that don't use forEach.

realityking:
Saves a few bytes when used in a browser.


mcollina:
CI: https://ci.nodejs.org/view/MyJobs/job/readable-stream-pipeline/10/

PR #344, opened 2018-07-03
WIP: readable-stream@3

### LINKED ISSUE #212: Remove isarray dependency

### LINKED ISSUE #283: destroy/pump
since pump relies on destroy, we probably want destroy in before we do pump.

### LINKED ISSUE #284: bumping the major
`readable-stream` v3!

There are a bunch of things we want to put in there that are semver-major, one of which is cutting down the supported node versions.

When do we want to do it? Node v9, Node v10?

### LINKED ISSUE #297: support the new error system from core
As we talked about it in May, core is implementing error codes, and `'stream'` should conform to this rule. The big question is how we port those to `readable-stream`, as I think it might be worthy to keep consistency.

I do not want a regexp for each error, so we might have to parse the error definition file and do some form or processing of that one as well.

cc @jasnell 

Example: https://github.com/nodejs/node/commit/d50a802feb0b061383400e3f290ddc7410cb3a48

mcollina:
As titled, this is a port of Node.js 10.5.0.
It includes all the semver-major changes of both Node 9 and 10.

Fixes #283 
Fixed #284 
Fixes #212 
Fixes #297


mcollina:
I've just released readable-stream@3.0.0-rc.1. Install it with `npm i readable-stream@next` and report bugs!

test/browser.js
```diff
@@ -11,6 +11,27 @@ if (!global.console.info) {
...
 var test = require('tape');
+var util = require('util');
+
+// TODO: add replacements instead
+if (!util.inspect) {
+  util.inspect = function () {};
+  util.inspect.custom = 'custom';
+}
+
+// TODO: add replacements instead
```
BridgeAR: Should these be replaced before release?

test/browser.js
```diff
@@ -49,8 +70,8 @@ test('streams 2', function (t) {
   require('./browser/test-stream2-pipe-error-once-listener')(t);
   require('./browser/test-stream2-push')(t);
   require('./browser/test-stream2-readable-empty-buffer-no-eof')(t);
-  require('./browser/test-stream2-readable-from-list')(t);
-  require('./browser/test-stream2-transform')(t);
+  // require('./browser/test-stream2-readable-from-list')(t);
+  // require('./browser/test-stream2-transform')(t);
```
BridgeAR: Dead code?

errors-browser.js
```diff
@@ -0,0 +1,61 @@
...
+createErrorType('ERR_INVALID_OPT_VALUE', function (name, value) {
+  return 'The value "' + value + '" is invalid for option "' + name + '"';
+}, TypeError);
+createErrorType('ERR_INVALID_ARG_TYPE', 'argument must be of the right type', TypeError);
+createErrorType('ERR_STREAM_PUSH_AFTER_EOF', 'stream.push() after EOF');
+createErrorType('ERR_METHOD_NOT_IMPLEMENTED', function (name) {
+  return 'The ' + name + ' method is not implemented';
+});
+createErrorType('ERR_STREAM_PREMATURE_CLOSE', 'premature close');
+createErrorType('ERR_STREAM_DESTROYED', 'the stream was destroyed');
```
BridgeAR: We used capital letters in Node.js for the beginning of the message.

In core `ERR_STREAM_DESTROYED` has the message: `'Cannot call %s after a stream was destroyed'`

errors-browser.js
```diff
@@ -0,0 +1,61 @@
...
+createErrorType('ERR_MULTIPLE_CALLBACK', 'Callback called multiple times');
+createErrorType('ERR_STREAM_CANNOT_PIPE', 'Cannot pipe, not readable');
+createErrorType('ERR_STREAM_WRITE_AFTER_END', 'write after end');
+createErrorType('ERR_STREAM_NULL_VALUES', 'May not write null values to stream', TypeError);
+createErrorType('ERR_UNKNOWN_ENCODING', function (arg) {
+  return 'Unknown encoding: ' + arg;
+}, TypeError);
+createErrorType('ERR_STREAM_UNSHIFT_AFTER_END_EVENT', 'stream.unshift() after end event');
```
BridgeAR: The error messages slightly diverge in general from core.

phated:
On node 8.11.1, readable-stream crashes the process with `Error: Cannot find module 'babel-runtime/core-js/symbol'`

Do we really need that?  Why would it not be included?

errors-browser.js
mcollina: Node.js test suite passes without changing anything regarding error codes. Would you be able to send a PR against this branch that updates the error messages to the one of Node.js core? All of those were ported by hand :/.

phated:
working in my small test now

calvinmetcalf:
we probably want a changelog or a notice in the readme about what breaking changes are in v3 to make upgrading easier, otherwise LGTM

mcollina:
@calvinmetcalf done.

PR #352, opened 2018-08-11
Fix error messages

vweevers:
Addresses https://github.com/nodejs/readable-stream/pull/344#discussion_r201010673 and https://github.com/nodejs/readable-stream/pull/344#discussion_r201010948.

### DIFF
--- a/errors-browser.js
+++ b/errors-browser.js
@@ -1,5 +1,7 @@
 'use strict';
 
+var _typeof = typeof Symbol === "function" && typeof Symbol.iterator === "symbol" ? function (obj) { return typeof obj; } : function (obj) { return obj && typeof Symbol === "function" && obj.constructor === Symbol && obj !== Symbol.prototype ? "symbol" : typeof obj; };
+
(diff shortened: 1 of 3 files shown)


vweevers:
[Test failure](https://travis-ci.com/nodejs/readable-stream/jobs/139453283) is unrelated (it's the travis+npm timeout issue).

vweevers:
> Would you mind adding some sort of unit test under test/ours?

OK. Should I use tap, tape or assert? The test suite is a mixture of these.

BridgeAR:
@mcollina thinking about the changes: as long as they initially are aligned it should be fine. Testing these could only be done properly in Node core itself and at least adding tests for `ERR_INVALID_ARG_TYPE` would be redundant in core as that error is tested very often, including the error message.

BridgeAR:
@vweevers all stream related Node core tests run against this module as well and that is how the tests should be tested. Having these tests here mean both sides could still diverge without noticing it. They probably do not hurt either tough.

To me it's your call if you keep them or remove them, even tough I would remove them.

vweevers:
> Having these tests here mean both sides could still diverge without noticing it. 

I see what you mean.

@mcollina your call :)

PR #354, opened 2018-08-12
Removed process-nexttick-args as it's not used anymore

mcollina:
It's not used in `lib/`, just to check if tests pass everywhere.


vweevers:
A [search](https://github.com/nodejs/readable-stream/search?q=process-nextick-args&unscoped_q=process-nextick-args) shows it's still used?

mcollina:
@vweevers thanks for checking. I've removed it everywhere, let's see what CI says.

mcollina:
CI seems happy. @vweevers can you please check?

Issue #355, opened 2018-08-16
break in  BufferList.prototype[inspect.custom] = function () {


mitra42:
* **Version**: node 8.8.1 readable-stream 3.0.1 
* **Platform**: Webpacked on node 8.8.1 webpack 4.16.5 running on current Firefox & Chrome

A change to readable-stream 3.0.1 from the previous 2.3.6 causes the following bug while loading the webpacked application on a browser, the application then fails to load. 

A simple rollback to 2.3.6 fixes this, so it looks like something broken in the major release.
```
buffer_list.js:155 Uncaught TypeError: Cannot read property 'custom' of undefined
    at eval (buffer_list.js:155)
    at eval (buffer_list.js:161)
    at Object../node_modules/readable-stream/lib/internal/streams/buffer_list.js (dweb-transports-bundle.js:16564)
    at __webpack_require__ (dweb-transports-bundle.js:20)
    at Object.eval (VM3191 _stream_readable.js:63)
    at eval (VM3191 _stream_readable.js:1038)
    at Object../node_modules/readable-stream/lib/_stream_readable.js (dweb-transports-bundle.js:16516)
    at __webpack_require__ (dweb-transports-bundle.js:20)
    at eval (VM3190 readable-browser.js:1)
    at Object../node_modules/readable-stream/readable-browser.js (dweb-transports-bundle.js:16610)
```
The line in question is break in  `BufferList.prototype[inspect.custom] = function () { `
And `inspect` is defined in:
```
var _require = __webpack_require__(/*! util */ 1),
    inspect = _require.inspect;
```
I thought it might be just an undeclared dependency on `inspect`  which I didn't find anywhere in package-lock.json, but even installing `inspect` (version 0.0.2) doesnt fix it, could it be an unforseen bad interaction with `webpack` ?

mcollina:
Thanks for reporting this!

It is the latter, I think their Node compatibility layer has not been updated in a bit of time.

On a side note, I think we should add some tests that verifies webpack compatibility.

Would you mind adding an example so I can verify the fix?

mitra42:
I'm not sure how to add a simple example for your test ... haven't worked on readable-streams (just using it in @internetarchive/dweb-transports).

Issue #356, opened 2018-08-16
Remove use of require('util') in lib/


mcollina:
Because it's hard for browsers.

mcollina:
See https://github.com/nodejs/readable-stream/pull/357#discussion_r210598132 for more details.

catamphetamine:
So, in the main file you have:
https://unpkg.com/readable-stream@3.0.6/readable.js

```js
var Stream = require('stream');
if (process.env.READABLE_STREAM === 'disable' && Stream) {
  module.exports = Stream;
  exports = module.exports = Stream.Readable;
  exports.Readable = Stream.Readable;
  exports.Writable = Stream.Writable;
  exports.Duplex = Stream.Duplex;
  exports.Transform = Stream.Transform;
  exports.PassThrough = Stream.PassThrough;
  exports.Stream = Stream;
} else {
  exports = module.exports = require('./lib/_stream_readable.js');
  exports.Stream = Stream || exports;
  exports.Readable = exports;
  exports.Writable = require('./lib/_stream_writable.js');
  exports.Duplex = require('./lib/_stream_duplex.js');
  exports.Transform = require('./lib/_stream_transform.js');
  exports.PassThrough = require('./lib/_stream_passthrough.js');
  exports.finished = require('./lib/internal/streams/end-of-stream.js');
  exports.pipeline = require('./lib/internal/streams/pipeline.js');
}
```

First, how is it supposed to not throw in a browser, because there's no "stream" package there.
Second, suppose it doesn't throw at `require('stream')` and executes the `else` statement.
Then it goes to `./lib/_stream_readable.js` and sees this:

```js
/*<replacement>*/
var debugUtil = require('util');
var debug = void 0;
if (debugUtil && debugUtil.debuglog) {
  debug = debugUtil.debuglog('stream');
} else {
  debug = function () {};
}
/*</replacement>*/
```

And now it throws because there's no `'util'` module in a browser.
https://github.com/catamphetamine/read-excel-file/issues/12

How does this all work in browsers?
Are you relying on bundlers to automatically substitute things like `require('util')` and `require('stream')` with their own magic?

mcollina:
This project always relied on bundlers.

danm:
I'd be happy to pick this up @mcollina are there any updates before i get started?

mcollina:
Nope. You'll need to work in https://github.com/nodejs/readable-stream/tree/master/build so when pulling from core the util dependencies are replaced correctly.

[...]

PR #357, opened 2018-08-16
Import Node 10.9.0. Redo the fix for util.inspect.custom

mcollina:
Fixes https://github.com/nodejs/readable-stream/issues/355

### DIFF
--- a/build/files.js
+++ b/build/files.js
@@ -180,10 +180,6 @@ function CorkedRequest(state) {
       /Object\.getPrototypeOf\((chunk)\) !== Buffer\.prototype/g
     , '!Buffer.isBuffer($1)'
   ]
-  , fixCopyBuffer = [
-      /Buffer\.prototype\.copy\.call\(src, target, offset\);/
-    , 'src.copy(target, offset);'
-  ]
   , errorsOneLevel = [
         /internal\/errors/
     ,   '../errors'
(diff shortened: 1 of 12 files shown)


lib/internal/streams/buffer_list.js
```diff
@@ -2,14 +2,16 @@
...
-var util = require('util');
+var _require = require('buffer'),
+    Buffer = _require.Buffer;
 
-var _require = require('util'),
-    inspect = _require.inspect;
+var _require2 = require('util'),
+    inspect = _require2.inspect;
+
+var custom = inspect && custom || Symbol('useless');
```
BridgeAR: Instead of using `Symbol('useless')` fall back to `'inspect'`. That way old Node versions should work properly.

vweevers:
Is this still needed?

https://github.com/nodejs/readable-stream/blob/2149b0dfecfd3af25dc85662daed8065a770b631/test/browser.js#L16-L20

build/files.js
```diff
@@ -280,24 +276,15 @@ module.exports['_stream_writable.js'] = [
 
 module.exports['internal/streams/buffer_list.js'] = [
     [
-      /(?:var|const) (?:{ )Buffer(?: }) = require\('buffer'\)(?:\.Buffer)?;/,
+      /const \{ inspect \} = require\('util'\);/,
       `
-const Buffer = require('buffer').Buffer
-const util = require('util')
+const { inspect } = require('util')
+const custom = inspect && inspect.custom || 'inspect'
```
mafintosh: Won't this browserify all of `util`? I wrote https://github.com/mafintosh/inspect-custom-symbol to get around that (util is pretty big!)

mafintosh:
Added a comment

build/files.js
mcollina: @mafintosh that module uses the symbol instead of `'inspect'` as a key. See https://github.com/mafintosh/inspect-custom-symbol/blob/master/index.js#L4. @BridgeAR should we use the symbol or the property here?

mcollina: Also see #356 to remove the use of util.

BridgeAR: It should be `'inspect'`. Otherwise old Node.js versions won't work and a symbol has no other effect on browserify in this case than the string.

I don't use browserify often enough to know exactly what it would pull in but it should only be the parts that are used and in this case it is only the inspection property. So that should be fine.

mcollina: I think we should land this as is and deal with removing util in #356. What do you think? May I get a couple of LGTM?

mafintosh: @BridgeAR it pulls in the entire `util` source. that's why people use the inherits module instead of util.inherits on npm also :)

@mcollina Ya, solving it elsewhere is fine by me

Issue #358, opened 2018-08-28
Move to Buffer.from()


gabrielschulhof:
`new Buffer(...)` is being used in [a bunch of places](https://github.com/nodejs/readable-stream/search?p=2&q=%22new+Buffer%22&unscoped_q=%22new+Buffer%22) even though it's deprecated. Can we move to `Buffer.from(...)` in those places?

mcollina:
Those are just browsers tests. So, yes we __should__ move to `Buffer.from()` and it's not even a semver-minor change. Would you like to send a PR?

gabrielschulhof:
Can do.

gabrielschulhof:
I guess this landed in 32c1045 so it can be closed, right?

PR #359, opened 2018-08-29
remove Buffer constructor usage (#358)

gabrielschulhof:
* Replace occurrences of `new Buffer(...)` with `Buffer.alloc(...)` or
  `Buffer.from(...)`, as appropriate


gabrielschulhof:
@mcollina IIUC I have removed the block.

examples/CAPSLOCKTYPER.JS
```diff
@@ -11,7 +11,7 @@ function MyStream () {
 inherits(MyStream, Transform);
 
 MyStream.prototype._transform = function (chunk, outputFn, callback) {
-  outputFn(new Buffer(String(chunk).toUpperCase()));
+  outputFn(Buffer.alloc(String(chunk).toUpperCase()));
```
vweevers: Shouldn't this be `.from`?

test/browser/test-stream-readable-event.js
```diff
@@ -16,7 +16,7 @@ function first(t) {
   };
 
   // This triggers a 'readable' event, which is lost.
-  r.push(new Buffer('blerg'));
+  r.push(Buffer.alloc('blerg'));
```
vweevers: Same (also below)

examples/CAPSLOCKTYPER.JS
mcollina: ouch, tests ere passing!

mcollina: @gabrielschulhof would you mind sending a follow-up PR?

gabrielschulhof:
Can do.

PR #360, opened 2018-08-30
replace Buffer.alloc() with Buffer.from() where appropriate

gabrielschulhof:
FWIW tests still pass locally.


vweevers:
It's weird that the tests were passing, because `.alloc('string')` is supposed to throw: https://github.com/nodejs/node/blob/03b825811ed4af0addcdf6e75bacb3dc1c4c5940/lib/buffer.js#L279

gabrielschulhof:
For sure ... 🤷

mcollina:
@vweevers those are only browser tests.

PR #389, opened 2018-12-11
fix isInteger for IE11

### LINKED ISSUE #390: Number.isInteger in IE11
I'm having this error when importing and using `readable-stream` in IE11
```
TypeError: Object doesn't support property or method 'isInteger'
       at getHighWaterMark (eval code:12:5)
```

It seems like `isInteger` in only used in `/lib/internal/streams/state.js` and can be fixed during build step

Here is a PR #389 fixing it

JMounier:
Fixes usage of `Number.isInteger` in `lib/internal/streams/state.js` 
```
TypeError: Object doesn't support property or method 'isInteger'
       at getHighWaterMark (eval code:12:5)
```

+ add new `integerIE11` rule in `build/files.js`
+ add dev dependency to `babel-plugin-transform-optional-catch-binding` (so that anyone can run ./build/build.js)

Fix #390


package.json
```diff
@@ -29,6 +29,7 @@
     "babel-plugin-transform-es2015-spread": "^6.22.0",
     "babel-plugin-transform-es2015-template-literals": "^6.8.0",
     "babel-plugin-transform-inline-imports-commonjs": "^1.2.0",
+    "babel-plugin-transform-optional-catch-binding": "^7.0.0-alpha.16",
```
mcollina: I'm not really confident in depending on alpha code. Is there a stable version of this transform?

JMounier: I don't see any stable version on npm and the latest one is already one year old.

Anyway this package is only necessary to build `readable-stream` from scratch. And so is not relevant to the fix.

I'll just drop the commit adding the dependency and create an issue (#392) to upgrade to babel 7 stable as @targos said.

Issue #408, opened 2019-05-09
readable-browser.js does not export everything readable.js does


angelo-v:
I am wondering, why `readable-browser.js` is missing some exports. E.g. `readable.js` is exporting `finished` and a library that I am using imports it. `readable-browser.js` does not export finished. Now my application breaks, because `finished` is undefined in a browser application.

Is this intentional? Am I using it wrong?

angelo-v:
`finished` and `pipeline` got added to `readable.js` in [this commit](https://github.com/nodejs/readable-stream/commit/74193c5acd78e17bac61b785ff32c0903833cd16#diff-ab6bca17203019541ae18ae63a2604d3) and it seams to me, that it just was forgotten to update `readable-browser.js`

mcollina:
Good spot! Would you like to send a PR?

angelo-v:
I could add those exports to readable-browser.js, yes. But what is the purpose to differ between `readable-browser.js` and `readable.js` in general? Is the separation still justified?

mcollina:
`readable.js` accesses the `process.env` object. `readable-browser.js` does not, and it will yield smaller bundles.
Overall, yes, they are justified and needed.

angelo-v:
Ok, I will prepare a PR

angelo-v:
fixed via https://github.com/nodejs/readable-stream/pull/409

PR #409, opened 2019-05-15
add missing exports for browsers

angelo-v:
This PR adds the missing exports, mentioned in https://github.com/nodejs/readable-stream/issues/408

There should be tests, that would have catched the problem beforehand, but I did not figure out how to add those in reasonable time. I appreciate, if someone more experienced with this project adds those tests.


mcollina:
Can you add/port a couple of test to our browser suite https://github.com/nodejs/readable-stream/tree/master/test/browser? As an example: https://github.com/nodejs/readable-stream/blob/master/test/parallel/test-stream-pipeline.js and https://github.com/nodejs/readable-stream/blob/master/test/parallel/test-stream-finished.js.

angelo-v:
I tried to do that, but failed. Cannot invest more time on that, sorry

angelo-v:
Anyone else that could do it? Would be cool to have this merged.

PR #410, opened 2019-05-27
Add node v12 to .travis.yml

mcollina:


package.json
```diff
@@ -27,14 +27,14 @@
     "nyc": "^11.0.0",
     "pump": "^3.0.0",
     "rimraf": "^2.6.2",
-    "tap": "^11.0.0",
+    "tap": "^12.0.0",
```
vweevers: Is there a reason we can't update to latest (v14)?

mcollina: it does not support Node 6

PR #414, opened 2019-07-08
Check for existence of `process` object (fix for v3)

### LINKED ISSUE #412: v2: `process is not defined` error when bundled with webpack

th0r:
Fixes #412 for v3

**Notes:**
- Library was built using the following command: `./build/build.js 10.15.3`
- Some unrelated files were changed because of the updated Babel version (updated runtime helpers)

### DIFF
--- a/next-tick-browser.js
+++ b/next-tick-browser.js
@@ -1,8 +1,0 @@
-module.exports = nextTick;
-
-function nextTick(fn) {
-  var args = Array.prototype.slice.call(arguments, 1);
-  setTimeout(function () {
-    fn.apply(null, args);
-  }, 0);
-}
--- a/next-tick-browser.js
+++ b/next-tick-browser.js
@@ -1,0 +1,8 @@
+module.exports = nextTick;
+
+function nextTick(fn) {
+  var args = Array.prototype.slice.call(arguments, 1);
+  setTimeout(function () {
+    fn.apply(null, args);
+  }, 0);
+}
--- a/next-tick.js
+++ b/next-tick.js
@@ -1,0 +1,1 @@
+module.exports = process.nextTick;
--- a/next-tick.js
+++ b/next-tick.js
@@ -1,1 +1,5 @@
-module.exports = process.nextTick;
+module.exports = nextTick;
+
+function nextTick(...args) {
+  process.nextTick(...args);
+}
--- a/next-tick.js
+++ b/next-tick.js
@@ -1,5 +1,0 @@
-module.exports = nextTick;
-
-function nextTick(...args) {
-  process.nextTick(...args);
-}
--- a/README.md
+++ b/README.md
@@ -15,7 +15,7 @@ npm install --save readable-stream
 
 This package is a mirror of the streams implementations in Node.js.
 
-Full documentation may be found on the [Node.js website](https://nodejs.org/dist/v10.15.3/docs/api/stream.html).
+Full documentation may be found on the [Node.js website](https://nodejs.org/dist/v10.16.0/docs/api/stream.html).
 
 If you want to guarantee a stable streams base, regardless of what version of
 Node you, or the users of your libraries are using, use **readable-stream** *only* and avoid the *"stream"* module in Node-core, for background see [this blogpost](http://r.va.gg/2014/06/why-i-dont-use-nodes-core-stream-module.html).
--- a/build/files.js
+++ b/build/files.js
@@ -212,6 +212,25 @@ function CorkedRequest(state) {
           /const \{ once \} = require\('internal\/util'\);/
       ,  'function once(callback) { let called = false; return function(...args) { if (called) return; called = true; callback(...args); }; }'
   ]
+  , nextTickUsage = [
+          /process\.nextTick\((.+?)\)/g
+      ,   'nextTick($1)'
+  ]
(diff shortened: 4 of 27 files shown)


next-tick-browser.js
```diff
@@ -0,0 +1,8 @@
+module.exports = nextTick;
+
+function nextTick(fn) {
+  var args = Array.prototype.slice.call(arguments, 1);
+  setTimeout(function () {
+    fn.apply(null, args);
+  }, 0);
+}
```
vweevers: This is non-breaking for webpack users, right? But potentially breaking for `browserify` users.

th0r: > This is non-breaking for webpack users, right?

Yes, it's a slightly modified version of the webpack's polyfill.

> But potentially breaking for `browserify` users.

Well, potentially yes, but I don't think they will face any bugs. The only difference between these implementations is that the browserify's version drains queue synchronously on subsequent calls to `nextTick`.

next-tick.js
```diff
@@ -0,0 +1,5 @@
+module.exports = nextTick;
```
mcollina:
```suggestion
module.exports = process.nextTick;
```

next-tick-browser.js
mcollina: have you got a link to Browserify impl? I think Browserify version might be slightly faster and provide a better experience for users overall.

next-tick.js
th0r: In this case it doesn't pass [this test](https://github.com/nodejs/readable-stream/blob/master/test/ours/lolex-fake-timers.js).

next-tick-browser.js
th0r: https://github.com/defunctzombie/node-process/blob/master/browser.js#L90-L154

next-tick.js
mcollina: the best approach is to use `module.exports.nextTick = process.nextTick` pattern instead. Adding a wrapper is going to slow things down.

next-tick-browser.js
mcollina: I prefer the browserify one, and more importantly this was designed to work with browserify first.

next-tick.js
th0r: As i told you above it won't pass [this test](https://github.com/nodejs/readable-stream/blob/master/test/ours/lolex-fake-timers.js) if I use your approach.

mcollina: Oh, I understand now. You can do the following on the node side:

```js
module.exports = process
```

While on the browser side:


```js
module.exports = { nextTick }
```

Then, in code:

```js
var wrap = require('./nextTick')
wrap.nextTick(function () {
  // this will work via lolex
})
``

th0r: Well, it's basically a polyfilling of a `process` object which I suggested here: https://github.com/nodejs/readable-stream/pull/413#issuecomment-508768361

mcollina: I think we were in agreement in there. I'll just not call it `process`.

th0r: But how do you want it to be called? `wrap`?

mcollina: We should not expect to be a full `process` object. In the browser, it won't be. It's an object with a `nextTick` property. we can call how we want "lightProcess" "fakeProcess" "nextTickWrap". This should enable that lolex support.

next-tick-browser.js
th0r: `test/browser/test-stream2-unpipe-drain.js` fails if I use browserify's version of `nextTick`.

I think it happens because test runner uses it's own version of `process.nextTick` polyfill (which comes from browserify) and `readable-stream` uses our own version so we end up with two different queues of async events.

webpack's version doesn't have this problem.

th0r: I've pushed a version with browserify's `nextTick`. If you want, you can try to fix this test - I don't have any more time to spend on it.

mcollina:
Can you please enable me to push on this PR? I have the test fixed.

th0r:
@mcollina I've added you as collaborator to the forked repo if this is what you are asking for.

th0r:
@mcollina any updates?

vweevers:
I'm -1 on the `nextTick` change.

- Maintaining a polyfill as inline code is painful
- This change, intended for webpack users, will also affect browserify users. I would prefer that webpack 5 users include `node-libs-browser` in their configuration to get webpack 4 behavior.
- I'm worried that having a separate nextTick queue will introduce subtle, hard-to-debug race issues.

PR #417, opened 2019-09-20
modify write method for Uint8Array output

tomcat008:
When I use MQTT.js in wechat and publish ArrayBuffer message，the subclient cannot recv the message.
I debugged the program and found in _stream_writable.js "write" method the ArrayBuffer  payload encoding is converted to “utf-8”,cannot convert to Buffer。
I've made changes and passed the test in the project.

### DIFF
--- a/lib/_stream_writable.js
+++ b/lib/_stream_writable.js
@@ -285,9 +285,15 @@ Writable.prototype.write = function (chunk, encoding, cb) {
   var state = this._writableState;
   var ret = false;
 
-  var isBuf = !state.objectMode && _isUint8Array(chunk);
+  // var isBuf = !state.objectMode && _isUint8Array(chunk);
 
-  if (isBuf && !Buffer.isBuffer(chunk)) {
+  // if (isBuf && !Buffer.isBuffer(chunk)) {
+  //   chunk = _uint8ArrayToBuffer(chunk);
+  // }
+
+  var isBuf = _isUint8Array(chunk);
+
+  if (isBuf) {
     chunk = _uint8ArrayToBuffer(chunk);
   }
 


lib/_stream_writable.js
```diff
@@ -285,9 +285,15 @@ Writable.prototype.write = function (chunk, encoding, cb) {
   var state = this._writableState;
   var ret = false;
 
-  var isBuf = !state.objectMode && _isUint8Array(chunk);
```
rexagod: Streams not operating under `objectMode` deal solely with Buffers, UInt8Arrays and Strings. IMO this check (`!state.objectMode`) is necessary and should be valid for all cases.

rexagod: Shouldn't `isBuf` be renamed to `isNotBuf` since `chunk` is a UInt8Array?

lib/_stream_writable.js
```diff
@@ -285,9 +285,15 @@ Writable.prototype.write = function (chunk, encoding, cb) {
...
+  // var isBuf = !state.objectMode && _isUint8Array(chunk);
 
-  if (isBuf && !Buffer.isBuffer(chunk)) {
```
rexagod: Similarly, I see no reason why `!Buffer.isBuffer(chunk)` check should be removed for a UInt8Array?

rexagod:
ping @tomcat008

vweevers:
The current code works as intended. In objectMode, the input is taken as-is (and we cannot change that as @rexagod wrote). If not in objectMode, then UInt8Arrays are converted to Buffers. That does not include ArrayBuffers though; if you want that to be supported (I'm not sure we should) please open an issue in [node](https://github.com/nodejs/node). Thanks!

PR #418, opened 2019-09-26
tests: add GitHub Actions CI

bnb:
PR to attempt adding GitHub Actions CI as a parallel to/eventual (?) replacement for TravisCI.

Varied a bit from the setups I've done so far, so I don't fully expect this to work on the first run. That said, I'll be working to get it passing.

Few things to note:

- There's not currently (AFAIK) a Sauce Labs Action that parallels what TravisCI currently does. I reached out on Twitter to ask if something existed or was planned, and @christian-bromann started working on an Action. Ideally we'd be able to use that once it's done ❤️ 
- The configuration to parallel `CMD=test` and `CMD=test-browser` is relatively straightforward and documented [here](https://help.github.com/en/articles/workflow-syntax-for-github-actions#jobsjob_idstrategymatrix), but omitted from this PR since the PR does not currently include Sauce Labs.
- I opted to run on all platforms since that's a personal preference of mine, but if there are platforms you'd prefer to be removed I'm happy to do so.


bnb:
Hmm, theoretically the Checks should have run, though I could (?) be wrong about that. Double checking in a dummy repo.

.github/workflows/ci.yml
```diff
@@ -0,0 +1,19 @@
+name: GitHub Actions CI
+
+on: [push]
```
targos:
```suggestion
on: [push, pull_request]
```

This might be why the check didn't run (the branch is on your fork, so pushes there won't trigger the push event here)

vweevers:
What's the motivation for replacing Travis?

bnb:
Per my understanding from a Build WG member when I asked about GitHub CI, it was agreed to attempt using GitHub CI in a few repos in the org that already use TravisCI. If it works as a reasonable replacement, it could eventually become that but for now perhaps not the goal we'd be going for.

.github/workflows/ci.yml
bnb: Doesn't seem to have actually done it – if I'd done it in a branch, I believe it would have worked. I imagine this is a sort of spam protection on their end – if I automate opening a bajillion PRs to FOSS repos with CI, that could be a massive drain on their infra 😅

bnb:
Looking into how I can trigger this correctly. Will update as appropriate ❤️

bnb:
Ah, it looks like they're running [on my repo](https://github.com/bnb/readable-stream/actions), it's just not being surfaced here! Excellent.

There's a failure with `tap` not being available, I'm realizing that's likely caused because I removed `npm install`. Will address and progress can be followed there ❤️

bnb:
Passed!

<details>
image
</details>

It's worth noting that this GitHub Actions CI build was triggered 6 minutes ago (from time of comment) and already completed on all operating systems on all versions, while TravisCI is waiting to be queued 😬

<details>
image
</details>

bnb:
Also, I noticed that the Travis run only tests Linux. I'd be happy to remove `macos-latst` and `windows-latest` if that'd be the preferred path of the maintainers.

.github/workflows/ci.yml
mcollina: @bnb a good feedback for the actions team is to put "pull request" in as default.

.github/workflows/ci.yml
```diff
@@ -0,0 +1,21 @@
+name: GitHub Actions CI
+
+on: [push, pull_request]
+
```
XhmikosR: I'd add `CI: true` here

.github/workflows/ci.yml
```diff
@@ -0,0 +1,21 @@
...
+        os: [ubuntu-latest, windows-latest, macOS-latest]
+        node-version: [6.x, 8.x, 10.x, 12.x]
+    steps:
+    - uses: actions/checkout@v1
+    - name: Use Node.js ${{ matrix.node-version }} on ${{ matrix.os }}
+      uses: actions/setup-node@v1
+      with:
+        node-version: ${{ matrix.node-version }}
+    - name: npm install and npm run test
+      run: |
```
XhmikosR: I'd split these 2

.github/workflows/ci.yml
```diff
@@ -0,0 +1,21 @@
+name: GitHub Actions CI
+
+on: [push, pull_request]
+
+jobs:
+  build:
+    runs-on: ${{ matrix.os }}
+    strategy:
```
XhmikosR: I'd add `fail-fast: false`

.github/workflows/ci.yml
XhmikosR:
```suggestion
env:
  CI: true

```

.github/workflows/ci.yml
XhmikosR:
```suggestion
    strategy:
      fail-fast: false
```

.github/workflows/ci.yml
mcollina: Isn't this set automatically?

XhmikosR: Nope.

bnb: Is this environment variable explicitly needed for this CI run? I generally try to avoid setting unnecessary environment variables, and it wasn't defined in the existing `.travis.yml`.

XhmikosR: It certainly does not hurt and it's what you had on Travis CI by default before.

XhmikosR: And in most if not all CI providers :P

bnb: If collaborators would like this added, I'd be happy to have it committed. Still cautious about adding additional environment variables that may be unnecessary.

XhmikosR: Well, it's not unnecessary. Many modules use this var to skip installing (like precommit packages) or showing messages on postinstall. And you had it already there (maybe without knowing).

bnb:
>Are tests passing already on Windows?

Seems like it 👍

.github/workflows/ci.yml
vweevers: Aye. `airtap` also uses it.

mcollina: Let’s add it then!

vweevers:
Has anyone evaluated the security of GH Actions? For example when tests needs secrets, like Sauce Labs tokens. To my understanding you must explicitly pass a secret to an action, other actions can't read it. That's good. Any other risks? Are actions mutable?

.github/workflows/ci.yml
vweevers: Not (or no longer) needed: https://docs.github.com/en/actions/reference/environment-variables#default-environment-variables

vweevers (approved):
LGTM. Regarding Sauce Labs, I can open a PR to create a separate workflow with an upgraded `airtap` (which now supports GH Actions).

christian-bromann:
> * There's not currently (AFAIK) a Sauce Labs Action that parallels what TravisCI currently does. I reached out on Twitter to ask if something existed or was planned, and @christian-bromann started working on an Action. Ideally we'd be able to use that once it's done

The GitHub action is done: https://github.com/saucelabs/sauce-connect-action

vweevers:
@christian-bromann Nice, thanks! Though we no longer need it here.

PR #419, opened 2019-10-20
Fixing a TypeError bug when state is undefined

TahaBoulehmi:
I'm fixing a bug. Sometimes state is undefined and I get a Type Error:
"TypeError: Cannot read property 'corked' of undefined"


mcollina:
This is an export from Node core. It would be good to have a way to reproduce this so we can fix it upstream.

vweevers:
If it can still be reproduced please open an issue in [node](https://github.com/nodejs/node) with details. Thanks!

isbdnt:
It can be reproduced by clearing all require.cache
``` javascript
function clearRequireCache() {
    Object.keys(require.cache).forEach(function (key) {
            delete require.cache[key];
    });
}
```

PR #428, opened 2020-02-13
[Fix] babel's "loose mode" class transform enbrittles BufferList

ljharb:
```js
Object.defineProperty(Object.prototype, util.inspect.custom, {
  set(v) { throw 'this should not happen inside readable-stream'; }
});
```

With this change, I believe the output will use [[Define]] instead of [[Set]] for https://github.com/nodejs/node/blob/c101251a95cc82142bee4637f8db6cc360a06d82/lib/internal/streams/buffer_list.js#L167, and thus no longer fail when Object.prototype is modified.

### DIFF
--- a/.babelrc
+++ b/.babelrc
@@ -13,11 +13,5 @@
       ],
       "debug": true
     }]
-  ],
-  "plugins": [
-    // Included in the preset, but we can't specify per-plugin options there
-    ["@babel/plugin-transform-classes", {
-      "loose": true
-    }]
   ]
 }
--- a/lib/internal/streams/buffer_list.js
+++ b/lib/internal/streams/buffer_list.js
@@ -6,6 +6,12 @@ function _objectSpread(target) { for (var i = 1; i < arguments.length; i++) { va
 
 function _defineProperty(obj, key, value) { if (key in obj) { Object.defineProperty(obj, key, { value: value, enumerable: true, configurable: true, writable: true }); } else { obj[key] = value; } return obj; }
 
+function _classCallCheck(instance, Constructor) { if (!(instance instanceof Constructor)) { throw new TypeError("Cannot call a class as a function"); } }
+
+function _defineProperties(target, props) { for (var i = 0; i < props.length; i++) { var descriptor = props[i]; descriptor.enumerable = descriptor.enumerable || false; descriptor.configurable = true; if ("value" in descriptor) descriptor.writable = true; Object.defineProperty(target, descriptor.key, descriptor); } }
+
+function _createClass(Constructor, protoProps, staticProps) { if (protoProps) _defineProperties(Constructor.prototype, protoProps); if (staticProps) _defineProperties(Constructor, staticProps); return Constructor; }
+
 var _require = require('buffer'),
     Buffer = _require.Buffer;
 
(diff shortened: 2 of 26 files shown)


ljharb:
Seems like https://github.com/vweevers/readable-stream/commit/432d4b77439a8a825d7a330c2d16410e9ac15fbf is when it first was put into loose mode, in #299, due to #293. 

Does readable-stream still support pre-ES5 environments?

mcollina:
We support IE11 in readable-stream v3, while we still support pre-es5 environments in v2. This can land but not be backported.

mcollina:
Can you also regenerate the code?

```
cd build
node build.js v10.10.0
```

ljharb:
just v3 is fine; I’m mainly concerned with the version of this that’s vendored in core; that one doesn’t need to run Babel at all :-)

will update the build shortly

lib/internal/streams/buffer_list.js
```diff
@@ -22,170 +28,183 @@ module.exports =
...
+        customInspect: false
+      }));
     }
-
-    this.length -= c;
-    return ret;
-  } // Make sure the linked list only shows the minimal necessary information.
-  ;
-
-  _proto[custom] = function (_, options) {
```
ljharb: this was the vulnerability to Object.prototype modification

lib/internal/streams/buffer_list.js
```diff
@@ -22,170 +28,183 @@ module.exports =
...
+        ++c;
       }
 
-      ++c;
+      this.length -= c;
+      return ret;
+    } // Make sure the linked list only shows the minimal necessary information.
+
+  }, {
+    key: custom,
```
ljharb: this is the `custom` method

lib/internal/streams/buffer_list.js
```diff
@@ -22,170 +28,183 @@ module.exports =
...
-  };
-
-  _proto.join = function join(s) {
-    if (this.length === 0) return '';
-    var p = this.head;
-    var ret = '' + p.data;
-
-    while (p = p.next) {
-      ret += s + p.data;
+  _createClass(BufferList, [{
```
ljharb: createClass uses Define internally, not Set

.babelrc
```diff
@@ -13,11 +13,5 @@
       ],
       "debug": true
     }]
-  ],
-  "plugins": [
-    // Included in the preset, but we can't specify per-plugin options there
-    ["@babel/plugin-transform-classes", {
```
mcollina: Can you drop this dependency from package.json?

lib/internal/streams/buffer_list.js
mcollina: Great!

.babelrc
ljharb: oh sure. It’s already included via the env preset but I’ll remove it anyways

ljharb: actually, there never was an explicit dep here (oops), it just worked by accident because env brings it in.

mcollina: Ooops.

mcollina:
cc @vweevers what do you think?

vweevers:
👍

ljharb:
yay! what are the next steps to get this into core?

mcollina:
I think you'll have to wait for npm to do a release, and then update core to that. Or you can force a replacement directly.

ljharb:
done in https://github.com/npm/cli/issues/824 and https://github.com/nodejs/node/pull/31977

PR #443, opened 2020-09-19
Upgrade airtap and use GitHub Actions

vweevers:
- Upgrade airtap to v4 (see its [upgrade guide](https://github.com/airtap/airtap/blob/master/UPGRADING.md))
- Move Sauce Labs tests from Travis to GitHub Actions
- Use playwright for headless local testing

Complements #418.

### DIFF
--- a/.github/workflows/sauce.yml
+++ b/.github/workflows/sauce.yml
@@ -1,0 +1,24 @@
+name: Sauce Labs
+on: push
+jobs:
+  test:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v1
+      - name: Set up node
+        uses: actions/setup-node@v1
+        with:
+          node-version: 14
+      - name: Install
+        run: npm install
+        env:
+          # Download Sauce Connect binary now instead of on first run
+          SAUCE_CONNECT_DOWNLOAD_ON_INSTALL: true
+      - name: Add host
+        run: echo "127.0.0.1 airtap.local" | sudo tee -a /etc/hosts
+      - name: Test
+        run: npm run test-browsers
+        env:
+          SAUCE_USERNAME: ${{ secrets.SAUCE_USERNAME }}
+          SAUCE_ACCESS_KEY: ${{ secrets.SAUCE_ACCESS_KEY }}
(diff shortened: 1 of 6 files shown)


.github/workflows/sauce.yml
```diff
@@ -0,0 +1,24 @@
...
+        env:
+          # Download Sauce Connect binary now instead of on first run
+          SAUCE_CONNECT_DOWNLOAD_ON_INSTALL: true
+      - name: Add host
+        run: echo "127.0.0.1 airtap.local" | sudo tee -a /etc/hosts
+      - name: Test
+        run: npm run test-browsers
+        env:
+          SAUCE_USERNAME: ${{ secrets.SAUCE_USERNAME }}
+          SAUCE_ACCESS_KEY: ${{ secrets.SAUCE_ACCESS_KEY }}
```
vweevers, line 24: @mcollina Can you add these two secrets?

vweevers:
Travis fails on npm install because playwright (which has an install script) requires node 10. I'll remove playwright for now, unless someone has an idea for a workaround.

mcollina:
They keys are set up

vweevers:
Build failed:

> Error: 23 Sep 19:50:19 - Sauce Connect could not establish a connection.

Will see if I can replicate that on another repo that has an almost identical setup, and no sauce connect problems.

vweevers:
With debug output enabled, the problem [seems to be the credentials](https://github.com/nodejs/readable-stream/runs/1157278209?check_suite_focus=true#step:6:148):

```
ERROR: connecting to https://saucelabs.com/rest/v1/***/tunnels: HTTP response code indicated failure, reply: {"message": "Not authorized"}.
ERROR: can't reach https://saucelabs.com/rest/v1/***/tunnels, please check your username/access key, firewall and proxy settings.
```

@mcollina Could you please double-check the credentials? You can also try it locally:

```
SAUCE_USERNAME=xx SAUCE_ACCESS_KEY=xx DEBUG=airtap* npm run test-browsers
```

mcollina:
Thanks! Fixed and now the tests are passing!

mcollina:
Running `npm test-browser` I'm getting:

```
Hostname 'airtap.local' must resolve to 127.0.0.1. Please add an entry to your hosts file.
```

vweevers:
@mcollina That hostname error is expected. Some browsers don't route localhost traffic through the Sauce Connect tunnel. To work around that, airtap uses a custom hostname that defaults to `airtap.local` and should resolve to 127.0.0.1. Only needed for sauce, not for playwright and others.

The debug steps I added in the last commit should be removed, especially `./node_modules/sauce-connect-launcher/sc/sc-4.6.2-linux/bin/sc` which will break when a new Sauce Connect version is released by Sauce Labs.

mcollina:
oh gosh... can you send a fresh PR?

PR #444, opened 2020-09-22
Use the shared Travis config

### LINKED ISSUE #442: Use the shared travis config with auto-updates for node releases
tl;dr: I'd like to open a PR to replace the `node_js` section in the `.travis.yml` with an imported https://github.com/nodejs/ci-config-travis/blob/master/lts/gte-6.yml

Background: [Package Maintenance WG](https://github.com/nodejs/package-maintenance) has built a repository of [shared Travis CI imports](https://github.com/nodejs/ci-config-travis). The imports will be automatically updated as new Node.js versions get released.

Judging by what's inside the current `.travis.yml` here what's inside `engines`, I suspect that the `lts` upgrade policy is the right fit (it adds new versions as soon as they are released, never removes LTS versions, removes non-LTS versions when they reach EOL)? Or should it be `all` (to take the strict interpretation of what's inside `engines`)?

Is that OK?

dominykas:
Closes #442

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,23 +1,22 @@
 language: node_js
+
 notifications:
   email: false
-matrix:
+
+import:
+  - nodejs/ci-config-travis:lts/gte-6.yml
+
+env:
+  - CMD=test
+
+jobs:
   fast_finish: true
   include:
-  - node_js: 6
-    env: CMD=test
-  - node_js: 8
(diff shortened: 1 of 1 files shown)


mcollina:
We are really migrating to Github actions as soon as I can get to it, but this is good in the meanwhile.

PR #446, opened 2020-09-24
Cleanup CI

vweevers:
Remove debug steps from sauce workflow (#443), remove travis, and replace badges in readme.


mcollina (approved):
lgtm

Issue #426, opened 2020-01-15
global.uint8Array global not defined


rcollette:
I'm trying to pull this library into an Angular application (webpack).   

The use case is that I am trying to use PDFKit and blob-stream to create a PDF in the browser.

I override the "stream" module path in my tsconfig.json

```
"paths": {
      "pdfkit": [
        "node_modules/pdfkit/js/pdfkit.standalone"
      ],
      "stream": [
        "node_modules/readable-stream"
      ]
 }
```

I am getting the error

```
Error: Uncaught (in promise): ReferenceError: global is not defined ReferenceError: global is not defined at Object../node_modules/readable-stream/lib/_stream_readable.js
```

I try to work around that by adding in my index.html
```
<script>
    global = window;
</script>
```

But then I get the error:

```
Error: Uncaught (in promise): ReferenceError: process is not defined ReferenceError: process is not defined at Object../node_modules/readable-stream/experimentalWarning.js
```
Which seems to be related to:
https://github.com/nodejs/readable-stream/issues/412

What are the conditions required to use this library in a browser?

My specific environment is:
```
Package                           Version
-----------------------------------------------------------
@angular-devkit/architect         0.13.0
@angular-devkit/build-angular     0.13.0
@angular-devkit/build-optimizer   0.13.0
@angular-devkit/build-webpack     0.13.0
@angular-devkit/core              7.3.0
@angular-devkit/schematics        7.3.9
@angular/cli                      7.3.9
@ngtools/webpack                  7.3.0
@schematics/angular               7.3.9
@schematics/update                0.13.9
rxjs                              6.5.4
typescript                        3.2.4
webpack                           4.29.0
```

mcollina:
I think you'll need this option: https://webpack.js.org/configuration/node/

rcollette:
Thank you @mcollina .   

The web pack configuration is no longer directly available in Angular.  We are on Angular 7 for the time being and the final mechanism for customizing the the WebPack config is available in Angular 8
https://netbasal.com/customize-webpack-configuration-in-your-angular-application-d09683f6bd22

That configuration option also goes away in WebPack 5 as discussed on the #412 issue, so I suspect this will similarly be a problem going forward.

We're going to go with a server side implementation for the time being.

PR #447, opened 2020-10-15
Rename global to globalThis

andreymaznyak:
https://github.com/nodejs/readable-stream/issues/426
Fix issue global is not defined. (For IE need polyfills).
And it work on node >= 12


mcollina:
This library supports Node 6+ plus.

Issue #451, opened 2020-11-14
[Proposal]: Deno build


Soremwar:
One important feature that Deno's std library has is the NodeJS polyfill. This allows many third party libraries developed for Node to run in Deno without a lot of effort from developers in order to support both platforms. One great module missing from this compatibility layer however, is the Stream module. Even though I myself have put a lot of effort into making it work by porting it to native TypeScript code, I recognize the mantainability of that port is gonna be a concern in the future, and possible bug detection or modifications right now are based on trial and error mostly, due to the complexity handled by the internals of Streams

Luckily, thanks to some experimentation from @benjamingr and myself, we have determined that consuming `readable-stream` natively is indeed possible from Deno with a few thinkering to a bundle produced by esbuild and some minor changes to library code.

If you all agree I'll start to work in adding this kind of support for Deno right away. Thank you for your efforts, I look forward to adding a "Deno compatible" badge in the front page of this library :)

mcollina:
I think that would be a great addition. I would kindly ask to support this long-term as I have no experience with deno.

readable-stream is one of the most popular modules of the npm ecosystem, and it would indeed help in supporting both runtimes.

Soremwar:
Well Deno isn't that different from Node in terms of syntaxis. As long as the tests for the Deno build keep passing, the only thing to watch out for are the typings for the public interfaces of the library. I'd be happy to offer that kind of support for as long as its needed, but I think it would be a really trivial work to do.

mcollina:
There are no types here.

Soremwar:
Don't worry, I'll provide them. In fact I already have them

Soremwar:
One question, how are the tests for the streams generated? Do you copy them from the Node repo and ensure they run here? Can they be modified to use (let's say) a different testing library other than Node's native one?

PR #452, opened 2020-11-17
Build for Deno

Soremwar:
Discussion in #451 

This needs some polishing, namely:

- [ ] Enable tests to work based on the Deno build
- [ ] Await some of my PRs to be merged into Deno repo that fix some issues with their current Node polyfill
- [ ] Discuss method of publishing, and talk about mantainability for the future
- [ ] Document Deno polyfills
- [ ] Document Crossplatform variables and modules


Soremwar:
Anyone that can help me debug this fails? I tried to test them on Node 6.17.1 and they all passed

mcollina (changes requested):
The changes done inside `lib` must be done inside `build/` instead. There is a bunch of scripts there that takes code from Node.js core and rewrite it so it's agnostics. We could potentially use some of the same machinery to generate a deno build.

mcollina:
(Unfortunately I do not have time to debug the build right now, I'm going for some extended leave and I'll be back full time next year).

benjamingr:
@mcollina can you point me at the code that generates readable-stream from Node core? I'm not actually familiar with it :]?

targos:
@benjamingr https://github.com/nodejs/readable-stream/blob/master/build/build.js

Soremwar:
Someone knows what version of Node the last build of readable stream is using? The README points out to it being Node 10 but I can't really tell.

benjamingr:
Looking at the issue I think it's v10.18.1

Soremwar:
Thanks @benjamingr

mcollina:
This seems pretty amazing! Are there any tests for this?

Soremwar:
@mcollina Actually...that's the main reason why there's been so little activity in this since it was created. The build has been working for a while, but making the tests work has proven to be a major headache due to the differences between the Deno test suite and something like tap

More progress to follow though, I'm on vacations so I land put some time on this

mcollina:
You can also write a few high level tests. We run separate tests for the browser for this reason, see https://github.com/nodejs/readable-stream/tree/master/test/browser for more details.

mcollina:
FYI: check out https://github.com/nodejs/readable-stream/pull/454.

jimmywarting:
I doubt that Deno won't get much traction for node streams, they are mostly just focused on web standarder, meaning: whatwg streams. Node also recently got them.

I always try to avoid core builtin's when i develop anything for cross platform that runs in Deno, Browser and Node.
I never want to bundle the hole readable-stream into the browser, it cost too much, I always try to make my page score 100% in lighthouse
i got a own size limit to somewhere like < 100 kb when i develop a js library for browser and node:stream can easily break that threshold

Issue #439, opened 2020-05-27
readable-stream v4


mcollina:
I think it's time to start planning readable-stream v4.
Which version of Node.js should we pull from?

I would be ambitious and target v14.
I would also drop support for Internet Explorer, as it will go EOL later this year.

cc @ronag @vweevers @mafintosh

mcollina:
We also have some other things we should fix:

1. rollup support
2. webpack 5 support #435

ronag:
> I would be ambitious and target v14.

+1

> I would also drop support for Internet Explorer, as it will go EOL later this year.

+1

mafintosh:
SGTM

ljharb:
@mcollina i still maintain that it's bad for the ecosystem to try to "fix" what webpack 5 broke, and would prefer to ask webpack 5 users to enable the [workaround](https://github.com/nodejs/readable-stream/pull/435#issuecomment-627151596).

vweevers:
Thinking further ahead, could stream ultimately use `queueMicrotask` in both node and browsers? After Node.js 10 is EOL (in 2021). Then stream is guaranteed to behave the same in node and browsers.

ronag:
> Thinking further ahead, could stream ultimately use queueMicrotask in both node and browsers?

Isn't that a case of just using `Promise.resolve().then(cb)` instead of `process.nextTick(cb)` (I think are dropping IE?). There would be performance implications. Not sure whether those would be noticeable or not. Doing such a change would be semver-major. I'm unsure of the consequences since `Promise` has higher priority than `nextTick`.

vweevers:
There is a timing concern, yes, because the `queueMicrotask` queue is processed after the `nextTick` queue. That could pose a challenge in node core tests and specific stream implementations (especially the older ones that implement their own `.destroy()` etc).

ronag:
While I'm not against the idea. I'm concerned that this might be a bit overly ambitious right now and could stall.

mcollina:
I would not push that change to the ecosystem. This can break in oh-so-many unexpected ways.

I tend to agree with @ljharb. However it would likely be on us to develop some strategy to support webpack 5.

tpluscode:
ES modules, please 🙏

mcollina:
what about es modules?

tpluscode:
> what about es modules?

Replace commonjs with modules. 

It would be very nice if the package did not rely on built-ins and commonjs so that it's runs natively in browser

[...]

PR #454, opened 2021-01-07
Fix/remove node globals - rebased & fixing IE issue

martinheidegger:
Rebased version of #435 with [patch](https://github.com/hugomrdias/readable-stream/pull/1) to fix IE.


mcollina:
@vweevers @ljharb wdyt?

lib/internal/streams/next-tick-browser.js
```diff
@@ -0,0 +1,29 @@
+'use strict';
+
+module.exports = function (cb) { 
+    var args = new Array(arguments.length - 1);
```
ljharb, line 4: it’s very slow to use new Array; why is this not slicing arguments into an array?

ljharb (dismissed):
I think that this approach guarantees that most folks will have more than one nextTick shim in their bundle, which will inflate bundle sizes far more than just fixing webpack 5’s broken behavior.

I don’t understand why any change is being made here when webpack 5’s choice was to force *users* to deal with this. I also don’t see how it’s sustainable to try to force every package in the ecosystem to make similar changes, nor why this will be a benefit to the ecosystem to now have dozens of duplicate shims that are impossible to deduce (because they’re all potentially subtly different).

lib/internal/streams/next-tick-browser.js
martinheidegger: Indeed. I updated it for performance and push-rebased: https://github.com/nodejs/readable-stream/pull/454/files#diff-773895e366710aaad05c2daf697abc214727c9de1227114b5694878bd33322b2

martinheidegger:
> I think that this approach guarantees that most folks will have more than one nextTick shim in their bundle.

Hmm, would it be okay to use/require `queueMicrotask` with a fallback to `nextTick` until we drop support for Node 10? It is quite well established and there are polyfills? https://caniuse.com/mdn-api_windoworworkerglobalscope_queuemicrotask

ljharb:
I think that anything that's in node itself shouldn't be provided, and that it is the exclusive job of a *node module bundler* to provide shims for node builtins in the browser. In other words, this package can and should do whatever's needed to support the desired node versions, only. If webpack 5 has abdicated part of its only job, that shouldn't impose a cost on the entire javascript ecosystem (not just the part using webpack 5).

benjamingr:
@ljharb 

>  If webpack 5 has abdicated part of its only job, that shouldn't impose a cost on the entire javascript ecosystem (not just the part using webpack 5).

Webpack already did and the cost was already imposed. 

Now maintainers have two choices:
 - We can impose the cost on the users.
 - We can take the burden on ourselves and provide users with a better experience.

Given we already have a volunteer PR for #2 and Martin spent the work picking up the slack - why not take this code?

ljharb:
@benjamingr because webpack users - 5 or otherwise- aren’t the only, or the most important, set of users in the ecosystem. This PR imposes a cost to *everyone* (or potentially ever other bundler user, depending on how it’s authored), and that is not a reasonable thing to inflict.

martinheidegger:
I was actually trying to suggest option 4: use the cross-platform `queueMicrotask()` instead of a platform specific `process.nextTick` and supply fallbacks for old version that don't support `queueMicrotask` but that we want to be supported. For test sake I tried out on [this branch](https://github.com/nodejs/readable-stream/compare/master...martinheidegger:fix/remove-node-globals-3) but I [couldn't get the tests to work](couldn't get the tests to work) _(Would appreciate hints on how to fixe them)_.

vweevers:
@martinheidegger I raised [that idea earlier](https://github.com/nodejs/readable-stream/issues/439#issuecomment-635146547) and @mcollina responded:

> I would not push that change to the ecosystem. This can break in oh-so-many unexpected ways.

Can we try to prove that statement and then think about solutions? Because in packages where node builtins can safely be replaced with browser builtins, Webpack's change can indeed have a positive impact. That is not the case here - as far as we know now. Shims are a trap because they provide a better (build) user experience at first but hurt in the long run ([1](https://github.com/nodejs/readable-stream/pull/454#pullrequestreview-563564681), [2](https://github.com/nodejs/readable-stream/pull/435#issuecomment-640658183), [3](https://github.com/nodejs/readable-stream/pull/435#issuecomment-633649561), [4](https://github.com/nodejs/readable-stream/pull/414#issuecomment-520201719)). So I prefer to hold the line while we do explore long-term alternatives.

For example, one difference between node's `queueMicrotask` and `nextTick` is that the `queueMicrotask` queue is processed after the `nextTick` queue. What are the implications of that (in userland, beyond breaking very specific expectations in our tests)?

martinheidegger:
@mcollina also [mentioned](https://github.com/nodejs/readable-stream/issues/439#issuecomment-635176813) that it would be on _us_ to support Webpack 5. I think this current PR is the best possible solution.

ljharb:
Note that a viable strategy to support webpack 5 is to simply link to instructions of how to use ProvidePlugin to effectively revert webpack 5’s harmful change.

martinheidegger:
I agree with you, updating this is too much of a mess.

PR #399, opened 2019-02-15
export * from `stream` when `process.env.READABLE_STREAM === 'disable'`

trxcllnt:
Fixes https://github.com/nodejs/readable-stream/issues/398.

@mcollina I'm not sure how to test this other than creating a new script that does `READABLE_STREAM=disable npm test`. That approach seems wrong (unless this library is supposed to stay ahead of node's API), because I get the following two failures:

```
TypeError [ERR_INVALID_CALLBACK]: Callback must be a function
    at popCallback (internal/streams/pipeline.js:61:11)
    at pipeline (internal/streams/pipeline.js:66:20)
    at /home/ptaylor/dev/readable-stream/test/parallel/test-stream-pipeline.js:77:5
    at _tryBlock (/home/ptaylor/dev/readable-stream/node_modules/assert/assert.js:425:5)
    at _throws (/home/ptaylor/dev/readable-stream/node_modules/assert/assert.js:444:12)
    at Function.assert.throws (/home/ptaylor/dev/readable-stream/node_modules/assert/assert.js:474:3)
    at Object.<anonymous> (/home/ptaylor/dev/readable-stream/test/parallel/test-stream-pipeline.js:76:10)
    at Module._compile (internal/modules/cjs/loader.js:721:30)
    at Object.Module._extensions..js (internal/modules/cjs/loader.js:732:10)
    at Module.load (internal/modules/cjs/loader.js:620:32)
test/parallel/test-stream-pipe-unpipe-streams.js ...... 1/1
test/parallel/test-stream-pipe-without-listenerCount.js .. 1/1
test/parallel/test-stream-pipeline.js ................. 0/1
  not ok test/parallel/test-stream-pipeline.js
    timeout: 30000
    file: test/parallel/test-stream-pipeline.js
    command: /home/ptaylor/.nvm/versions/node/v11.6.0/bin/node
    args:
      - test/parallel/test-stream-pipeline.js
    stdio:
      - 0
      - pipe
      - 2
    cwd: /home/ptaylor/dev/readable-stream
    exitCode: 1

test/parallel/test-stream-readable-async-iterators.js .Mismatched noop function calls. Expected exactly 1, actual 0.
    at Object.mustCall (/home/ptaylor/dev/readable-stream/test/common/index.js:343:10)
    at Object.<anonymous> (/home/ptaylor/dev/readable-stream/test/parallel/test-stream-readable-async-iterators.js:647:21)
    at Module._compile (internal/modules/cjs/loader.js:721:30)
[...]

PR #457, opened 2021-07-07
Don't assign to readonly `promises` property when `READABLE_STREAM === 'disable'`

trxcllnt:
Follow-up to https://github.com/nodejs/readable-stream/pull/399. This PR fixes the below error, caused by recent versions of node adding a readonly `promises` property to the `Stream` constructor.

```
$ READABLE_STREAM=disable node
Welcome to Node.js v15.14.0.
Type ".help" for more information.
> var Readable = require('.')
Uncaught TypeError: Cannot set property promises of function Stream(opts) {
  EE.call(this, opts);
} which has only a getter
    at Function.assign (<anonymous>)
    at Object.<anonymous> (/home/ptaylor/dev/readable-stream/readable.js:4:10)
    at Module._compile (node:internal/modules/cjs/loader:1092:14)
    at Object.Module._extensions..js (node:internal/modules/cjs/loader:1121:10)
    at Module.load (node:internal/modules/cjs/loader:972:32)
    at Function.Module._load (node:internal/modules/cjs/loader:813:14)
    at Module.require (node:internal/modules/cjs/loader:996:19)
    at require (node:internal/modules/cjs/helpers:92:18)
> 
```

### DIFF
--- a/readable.js
+++ b/readable.js
@@ -3,7 +3,7 @@ if (process.env.READABLE_STREAM === 'disable' && Stream) {
   module.exports = Stream.Readable;
   Object.entries(Object.getOwnPropertyDescriptors(Stream))
     .filter((entry) => entry[1].writable && entry[0] !== 'prototype')
-    .forEach((entry) => module.exports[entry[0]] = Stream[entry[0]])
+    .forEach((entry) => module.exports[entry[0]] = Stream[entry[0]]);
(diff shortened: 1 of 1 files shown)


trxcllnt:
cc @mcollina because you helped review #399.

mcollina (approved):
lgtm

Issue #348, opened 2018-07-16
Circular dependencies (prevents proper `rollup`)


danielgindi:
Rollup spits those out:
```
Circular dependency: node_modules\readable-stream\lib\_stream_duplex.js -> node_modules\readable-stream\lib\_stream_readable.js -> node_modules\readable-stream\lib\_stream_duplex.js
Circular dependency: node_modules\readable-stream\lib\_stream_duplex.js -> node_modules\readable-stream\lib\_stream_readable.js ->  commonjs-proxy:C:\Users\Daniel.SILVERBOLT\Desktop\git\endpoint-analyzer\node_modules\readable-stream\lib\_stream_duplex.js -> nod
e_modules\readable-stream\lib\_stream_duplex.js
Circular dependency: node_modules\readable-stream\lib\_stream_writable.js -> node_modules\readable-stream\lib\_stream_duplex.js -> node_modules\readable-stream\lib\_stream_writable.js
Circular dependency: node_modules\readable-stream\lib\_stream_writable.js -> node_modules\readable-stream\lib\_stream_duplex.js ->  commonjs-proxy:C:\Users\Daniel.SILVERBOLT\Desktop\git\endpoint-analyzer\node_modules\readable-stream\lib\_stream_writable.js -> n
ode_modules\readable-stream\lib\_stream_writable.js
```

It is a huge problem, and will make code unexecutable after "rolling up".
The rollup process will convert commonjs deps into "static imports", which do not support circular dependencies.
Then it will end up with a flat hierarchy where there's a reference to a (yet) undefined variable.

mcollina:
Do you think this is a rollup problem or a readable-stream one? IMHO, it’s the former, as circular deps are valid cjs.

I’m happy to have a look anyway. Can you please produce a project/full example that demonstrate the issue?

danielgindi:
It's kind of both. The problem is not really rollup, as what rollup does is "convert" (through a plugin) into plain ES static imports, which do not support circular dependencies.

Circular dependencies are valid in cjs - but are still problematic even there because you have to be really careful about what is really export/imported and when does the circular dependency break it.
It's really a design flaw- that could be easily fixed by lifting common code/data into separate js files.

I'll try to make up a sample project

danielgindi:
Here is an example:

[rollup-example.zip](https://github.com/nodejs/readable-stream/files/2197507/rollup-example.zip)

[...]

PR #458, opened 2021-09-03
Fix circular dependency

eigendude:
## Description

This PR fixes a circular dependency that has been affecting the rollup ecosystem for a long time. The fix is simple, as the circularity comes from a single type test that can be simply replaced with a boolean. (And the boolean already exists, it's just exposed to the Duplex child class now).

The root cause is due to rollup's handling of hoisting. Duplex inherits from Readable and Writable, but Readable and Writable check for Duplex-ity, ergo circular. As a result, when Duplex inherits from Readable/Writable, these two classes are hoisted but not yet defined.

The fix is simple: Replace the Duplex-ity test with a boolean, set by Duplex when it constructs its base classes.

This boolean actually already exists locally in both Readable and Writable, so it makes sense to use it to bust the circular dependency.

## Related PRs

Closes https://github.com/nodejs/readable-stream/issues/348
Closes https://github.com/calvinmetcalf/rollup-plugin-node-builtins/issues/31

### DIFF
--- a/lib/_stream_duplex.js
+++ b/lib/_stream_duplex.js
@@ -57,8 +57,8 @@ require('inherits')(Duplex, Readable);
 
 function Duplex(options) {
   if (!(this instanceof Duplex)) return new Duplex(options);
-  Readable.call(this, options);
-  Writable.call(this, options);
+  Readable.call(this, options, true);
+  Writable.call(this, options, true);
   this.allowHalfOpen = true;
 
   if (options) {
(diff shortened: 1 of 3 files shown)


eigendude:
Is there any way to get this reviewed? Thanks.

fend25:
@mcollina hi!
Could you please review this PR? It is unbelievably helpful and long awaited. Thank you.

eigendude:
Is there a reason for the closure?

mcollina:
It's fixed in the upcoming v4 release.

eigendude:
Nice!!! Mind linking to the commit?

Also, I ported this patch to rollup-plugin-node-polyfills: https://github.com/ionic-team/rollup-plugin-node-polyfills/pull/36

PR #464, opened 2021-10-23
fix: remove reference to global (browser support)

### LINKED ISSUE #426: global.uint8Array global not defined

tgambet:
## Description
This simple PR is about making `readable-stream` compatible with browsers without needing a `global` polyfill.

Fixes #426 

## Discussion
Since global objects are available everywhere, I believe that `global.Uint8Array` can be safely replaced by `Uint8Array`.
This fixes some issues with browser support, particularly when using this library in a Web worker, or when shimming `global` is not possible.

### DIFF
--- a/lib/_stream_readable.js
+++ b/lib/_stream_readable.js
@@ -45,7 +45,7 @@ var Stream = require('./internal/streams/stream');
 
 var Buffer = require('buffer').Buffer;
 
-var OurUint8Array = global.Uint8Array || function () {};
+var OurUint8Array = Uint8Array || function () {};
 
 function _uint8ArrayToBuffer(chunk) {
   return Buffer.from(chunk);
--- a/lib/_stream_writable.js
+++ b/lib/_stream_writable.js
@@ -69,7 +69,7 @@ var Stream = require('./internal/streams/stream');
 
 var Buffer = require('buffer').Buffer;
 
-var OurUint8Array = global.Uint8Array || function () {};
+var OurUint8Array = Uint8Array || function () {};
 
 function _uint8ArrayToBuffer(chunk) {
   return Buffer.from(chunk);


lib/_stream_readable.js
```diff
@@ -45,7 +45,7 @@ var Stream = require('./internal/streams/stream');
 
 var Buffer = require('buffer').Buffer;
 
-var OurUint8Array = global.Uint8Array || function () {};
+var OurUint8Array = Uint8Array || function () {};
```
jimmywarting, line 48: this dose not quite work... you can't do `var x = foobar || function() {}`
you need to do: `var x = typeof foobar !== 'undefined' ? foobar : function() {}`

Checking if something exist globally must be done either via a global object like window, self, globalThis etc or using `typeof` checks like i suggested

tgambet: Got it, thanks for the explanation.

baybal:
@jimmywarting is the fix being merged?

Issue #473, opened 2022-05-30
process is not defined on SSR (Remix)


mzaatar:
Hi! 👋 
      
Firstly, thanks for your work on this project! 🙂

Today I used [patch-package](https://github.com/ds300/patch-package) to patch `readable-stream@3.6.0` for the project I'm working on.

I'm using this library (as a sub-dependency of [Terra Station wallet](https://docs.terra.money/docs/learn/terra-station/README.html) in a [Remix](https://remix.run/) and this library throws an exception on client-side `process is not defined`

Here is the diff that solved my problem (after polyfill `process` to client):

```diff
diff --git a/node_modules/readable-stream/lib/_stream_readable.js b/node_modules/readable-stream/lib/_stream_readable.js
index 192d451..244906c 100644
--- a/node_modules/readable-stream/lib/_stream_readable.js
+++ b/node_modules/readable-stream/lib/_stream_readable.js
@@ -45,6 +45,9 @@ var Stream = require('./internal/streams/stream');
 
 var Buffer = require('buffer').Buffer;
 
+var process = require('process');
+
 var OurUint8Array = global.Uint8Array || function () {};
 
 function _uint8ArrayToBuffer(chunk) {
diff --git a/node_modules/readable-stream/lib/_stream_writable.js b/node_modules/readable-stream/lib/_stream_writable.js
index a2634d7..7686a4b 100644
--- a/node_modules/readable-stream/lib/_stream_writable.js
+++ b/node_modules/readable-stream/lib/_stream_writable.js
@@ -69,6 +69,9 @@ var Stream = require('./internal/streams/stream');
 
 var Buffer = require('buffer').Buffer;
 
+var process = require('process');
+
 var OurUint8Array = global.Uint8Array || function () {};
 
 function _uint8ArrayToBuffer(chunk) {
```

<em>This issue body was [partially generated by patch-package](https://github.com/ds300/patch-package/issues/296).</em>

mcollina:
@ShogunPanda could you check if we will need this kind of hacks in v4 too?

ShogunPanda:
I would love to. @mzaatar As I'm not familiar with Remix, would you please share a minimum unpatched repo example I can play it?

ShogunPanda:
Ok, I've checked with a minimal repro. readable-stream v4 by definition will not affect the global scope and thus `process`, `Buffer` or similar will be unavailable.

Anyway, we have solution for all major bundlers, as we use them in our CI process.
[...]

PR #476, opened 2022-06-14
chore: Move playwright to optionalDependencies.

### LINKED ISSUE #475: Improve playwright harness
Currently:

1. it is not installing in any of my machines, the download get stuck forever
2. `npm install` get stuck and does not complete on CI (ref https://github.com/microsoft/playwright/issues/14853)

I think we should:

* [ ] split the integration / browser testing dependencies in its own `package.json` so we can avoid installing 113 MB
* [ ] ensure that the playwright "big download" is cached in CI so it is faster

@ShogunPanda could you take a look?

ShogunPanda:
Fixes #475 and #473.

### DIFF
--- a/package.json
+++ b/package.json
@@ -64,7 +64,6 @@
     "eslint-plugin-import": "^2.26.0",
     "eslint-plugin-n": "^15.2.0",
     "eslint-plugin-promise": "^6.0.0",
-    "playwright": "^1.21.1",
     "prettier": "^2.6.2",
     "process-es6": "^0.11.6",
     "rollup": "^2.72.1",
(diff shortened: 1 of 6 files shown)


mcollina:
Can you put a timeouton the browser workflows?

ShogunPanda:
@mcollina Done!

package.json
```diff
@@ -77,6 +76,9 @@
     "webpack": "^5.72.1",
     "webpack-cli": "^4.9.2"
   },
+  "optionalDependencies": {
```
vweevers, line 79: optionalDependencies are production dependencies (i.e. will be installed by dependents)

vweevers:
@mcollina I don't fully understand why you don't want to install playwright. Locally you'll only need to download the playwright binaries once, after that it's cached. Is it about CI? If so, you can selectively set `PLAYWRIGHT_SKIP_BROWSER_DOWNLOAD=1`

ShogunPanda:
@vweevers Yup, I think is to avoid to block CI in case those outages happen again. 

In general, since downloading browsers might be expensive (large files on mobile networks) I think we should not do it by default.

mcollina:
Maybe this is a bad idea and we can just skip this? Everything seems to be actually working fine now.

ShogunPanda:
Sounds good to me! Closing this!

PR #478, opened 2022-07-09
Lazily require abort-controller

hildjj:
It throws when it has been bundled, then loaded in node.

How I found this:

- Updating dependencies in [cbor](https://github.com/hildjj/node-cbor/)
- In the `cbor-web` sub-package, I use webpack to pre-bundle `cbor` for easy use in the browser
- I then spot-check that webpack'd version in node to ensure that it works as expected
- This error: `TypeError: Cannot destructure property 'AbortController'`
- Tracked it back to [code](https://github.com/mysticatea/abort-controller/blob/a935d38e09eb95d6b633a8c42fcceec9969e7b05/browser.js#L5) in `abort-controller`, which *should* only get used in the browser, but because I'm running a bundle, I hit it in node
- Backtracked into readable-stream.  There's no reason to require `abort-controller` unless `globalThis.AbortController` doesn't exist
- Moving the require after the `globalThis.AbortController` check fixes my issue

This is probably an edge case that I hope nobody else runs into, but the fix for my problem shouldn't hurt anyone else.

### DIFF
--- a/build/replacements.mjs
+++ b/build/replacements.mjs
@@ -14,7 +14,6 @@ const internalStreamsAbortControllerPolyfill = [
   `
   'use strict'
 
-  const abortControllerModule = require('abort-controller');
   `
 ]
 
@@ -30,7 +29,7 @@ const internalStreamsInspectCustom = ['inspect.custom', "Symbol.for('nodejs.util
 
 const internalStreamsNoRequireAbortController = [
   'const \\{ AbortController \\} = .+',
-  'const AbortController = globalThis.AbortController || abortControllerModule.AbortController;'
+  'const AbortController = globalThis.AbortController || require(\'abort-controller\').AbortController;'
 ]
 
 const internalStreamsRequireInternal = ["require\\('internal/([^']+)'\\)", "require('../$1')"]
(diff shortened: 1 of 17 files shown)


hildjj:
It looks like your `Browsers / Browsers (macos-latest, firefox, webpack)` test may have a timing issue.  I saw a failure of that test in my fork that went away when I re-ran that test.

mcollina:
Could you do the changes in the build script and re-run them? 

Take a look at https://github.com/nodejs/readable-stream/blob/main/build/replacements.mjs.

hildjj:
Things I didn't fix (yet) because they aren't germane to the issue at hand:

- Your build script appears to be using a bad URL template for node versions.  (https://nodejs.org/v18.0.0/node-v18.0.0.tar.gz instead of https://nodejs.org/dist/v18.0.0/node-v18.0.0.tar.gz)
- I tried using v18.5, but that broke a bunch of your tests
- Whitespace and semi-colon consistency changes introduced by the scripts.  I decided to leave what the script did, which I hope was correct. 
- Existing lint issues with `src/errors.js`

mcollina:
Good work!

PR #489, opened 2022-09-14
Remove requirement for bundler custom config

bergos:
The current code requires adding custom configs for bundlers to import Node.js polyfills. This PR removes that requirement.

Polyfills are required directly in the code. This can be done because the Node.js core modules have the same names (`buffer`, `events`, `process`), and therefore, the code modules are used under Node.js. The polyfills are only imported for browser builds.

This required switching from `buffer-es6` to `buffer`, and `process-es6` to `process`. Both packages are more up-to-date than the ones that have been replaced.

Many lines have been touched, but the `lib` folder changes are pretty small. Most changes are related to the test code. The same imports are required, but more files are affected.

Besides the globals `buffer` and `process`, `events` is added as a dependency. No code changes are required in that regard because `events` is not accessible as global, and therefore, `require` calls are already in the current code.

### DIFF
--- a/lib/internal/streams/destroy.js
+++ b/lib/internal/streams/destroy.js
@@ -1,4 +1,8 @@
-'use strict'
+/* replacement start */
+const process = require('process')
+/* replacement end */
+
+;('use strict')
 
 const {
   aggregateTwoErrors,
--- a/lib/internal/streams/destroy.js
+++ b/lib/internal/streams/destroy.js
@@ -1,8 +1,9 @@
+'use strict'
+
 /* replacement start */
+
 const process = require('process')
 /* replacement end */
-
-;('use strict')
 
 const {
   aggregateTwoErrors,
--- a/src/test/browser/fixtures/esbuild-browsers-shims.mjs
+++ b/src/test/browser/fixtures/esbuild-browsers-shims.mjs
@@ -1,8 +1,6 @@
-import * as bufferModule from 'buffer-es6'
-import * as processModule from 'process-es6'
+import * as processModule from 'process'
 
 export const process = processModule
-export const Buffer = bufferModule.Buffer
 
 export function setImmediate(fn, ...args) {
   setTimeout(() => fn(...args), 1)
--- a/build/build.mjs
+++ b/build/build.mjs
@@ -15,6 +15,7 @@ import { headers } from './headers.mjs'
 import { replacements } from './replacements.mjs'
 
 const baseMatcher = /^(?:lib|test)/
+const strictMatcher = /^(['"']use strict.+)/
 
 function highlightFile(file, color) {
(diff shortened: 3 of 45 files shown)


mcollina (changes requested):
Thanks for this PR! This change would need to be applied to our build/ scripts, not inside lib.

mcollina:
@ShogunPanda could you review?

lib/internal/streams/destroy.js
```diff
@@ -1,4 +1,8 @@
-'use strict'
+/* replacement start */
+const process = require('process')
+/* replacement end */
+
+;('use strict')
```
ShogunPanda, line 5: Can't you put require after the `use strict`?
This applies to this and all other similar files.

src/test/browser/fixtures/esbuild-browsers-shims.mjs
```diff
@@ -1,8 +1,6 @@
-import * as bufferModule from 'buffer-es6'
-import * as processModule from 'process-es6'
+import * as processModule from 'process'
 
 export const process = processModule
```
ShogunPanda, line 3: I'm a little confused here.
Can't you also include `process` in the headers requires as you did for `buffer` so that also that can be removed?

bergos: This is required by `tape` or a related package and can't be patched in the code.

lib/internal/streams/destroy.js
bergos: This is generated by the [headers.mjs](https://github.com/nodejs/readable-stream/blob/main/build/headers.mjs) build script. I agree that it is not very pretty, but there is existing code with the same patter: [test-stream-writable-destroy.js](https://github.com/nodejs/readable-stream/blob/main/test/parallel/test-stream-writable-destroy.js)

ShogunPanda: You could resolve this by replacing line 104 of `header.mjs` with this:

```javascript
if (content.match(strictMatcher)) {
  content = content.replace(strictMatcher, `$&;${header}`)
} else {
  content = header + content
}
```

The else branch was the original line.

Where: 

`const strictMatcher = /^(['"']use strict.+)/` 

is added below line 17.

ShogunPanda:
LGTM, if you could address https://github.com/nodejs/readable-stream/pull/489#discussion_r973996637 it would be great.

lib/internal/streams/destroy.js
bergos: @ShogunPanda thanks! I changed the build code. Now the patched code looks much prettier.

ShogunPanda (approved):
LGTM!

mcollina (approved):

Issue #487, opened 2022-09-08
cannot use readable-stream on service worker


userquin:
I patch these 2 modules with a postinstall script ("postinstall": "esno scripts/patch.ts"):
```ts
import { readFileSync, writeFileSync } from 'node:fs'

const patch = () => {
    let content: string = readFileSync(
        './node_modules/readable-stream/lib/_stream_readable.js',
        { encoding: 'utf-8' },
    )
    if (content.includes('global.Uint8Array')) {
        writeFileSync(
            './node_modules/readable-stream/lib/_stream_readable.js',
            content.replace(
                'var OurUint8Array = global.Uint8Array || function () {};',
                'var OurUint8Array = self.Uint8Array || function () {};'
            ),
            { encoding: 'utf-8' },
        )
    }
    content = readFileSync('./node_modules/readable-stream/lib/_stream_writable.js', { encoding: 'utf-8' })
    if (content.includes('global.Uint8Array')) {
        writeFileSync(
            './node_modules/readable-stream/lib/_stream_writable.js',
            content.replace(
                'var OurUint8Array = global.Uint8Array || function () {};',
                'var OurUint8Array = self.Uint8Array || function () {};'
            ),
            { encoding: 'utf-8' },
        )
    }
}

patch()
```

benmccann:
Rather than just assuming `global` will be defined, probably need to do something like:
```
typeof global !== "undefined"
  ? global
  : typeof self !== "undefined"
  ? self
  : typeof window !== "undefined"
  ? window
  : {};
```

E.g. as done in https://github.com/Aslemammad/modern-node-polyfills/blob/main/global.ts

mcollina:
Would you like to send a PR? Checkout our build scripts inside `build/`.

smeng9:
Mark as closed by https://github.com/nodejs/readable-stream/pull/502

PR #493, opened 2022-11-07
fix global undefined on service worker

smeng9:
Fixes https://github.com/nodejs/readable-stream/issues/487

### DIFF
--- a/build/replacements.mjs
+++ b/build/replacements.mjs
@@ -91,6 +91,8 @@ const internalValidatorsRequireRelativeUtil = ["require\\('internal/util'\\)", "
 
 const internalValidatorsRequireUtilTypes = ["require\\('internal/util/types'\\)", "require('../ours/util').types"]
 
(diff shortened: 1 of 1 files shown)


mcollina (approved):
lgtm

mcollina:
Do you think it would be possible to add a unit test under https://github.com/nodejs/readable-stream/tree/main/test/browser?

smeng9:
What kind of tests are we expected to run for this case? Do you mean add a runner-worker.mjs for service worker?

mcollina:
It should run a service worker and get back some data.

smeng9:
Hi @mcollina the worker setup seems is way more complicated than I thought.

Also this pull request should only target v3.6.0 https://github.com/nodejs/readable-stream/blob/bed7ffa274f5b9e6d0d5c22369e6fe825ded03d2/lib/_stream_readable.js where v4 seems does not have such issue.

Unfortunately, some of our dependencies still uses v3.

mcollina:
Can you target the v3.x line with your PR then? The build scripts were redone between them.

smeng9:
Hi @mcollina I cannot find the 3.x branch here https://github.com/nodejs/readable-stream/branches/stale?page=1 to target my commit with. Can you help to create a 3.x branch in this repo? Thanks!

Then I can fork it and create a merge.

Issue #482, opened 2022-08-20
`Readable.fromWeb` and `Readable.toWeb` do not seems to be properly implemented


Tpt:
First, thank you for your amazing work on `readable-stream`!

`Readable.fromWeb` an `Readable.toWeb` are calling the [`newStreamReadableFromReadableStream`](https://github.com/nodejs/readable-stream/blob/81c52641299e5ea3a5e37179967bb13691cad081/lib/internal/streams/readable.js#L1289) and [`newReadableStreamFromStreamReadable`](https://github.com/nodejs/readable-stream/blob/81c52641299e5ea3a5e37179967bb13691cad081/lib/internal/streams/readable.js#L1293) functions on the `webStreamsAdapters`.

However `webStreamsAdapters` [is only defined as an empty object](https://github.com/nodejs/readable-stream/blob/81c52641299e5ea3a5e37179967bb13691cad081/lib/internal/streams/readable.js#L1284) leading to a failure.

The same problem seems to affect `Writable` and `Duplex`.

mcollina:
Thanks for reporting! Which Node.js version are you using?
This module does not include WebStreams, so I won't expect those method to work... however we should likely throw a better exception.

Tpt:
> Thanks for reporting! Which Node.js version are you using?

NodeJS v18.7.0 so I would expect WebStreams to be available.

>  This module does not include WebStreams, so I won't expect those method to work... however we should likely throw a better exception.

What about having an implementation for cases like NodeJS 16+ and modern web browsers were WebStreams are available?

benjamingr:
@mcollina I think the ask is for these methods to work since the environment this package runs in (either Node to support several versions or browsers) often does have web streams?

mcollina:
I understand now, thanks!

Would you like to send a PR? I'll be a bit low on time for the coming months.

RishabhKodes:
Hi @mcollina @benjamingr, since this is an old issue, just wanted to know if this is still relevant or if has it been solved in any way, if not I can work on this. 

Had a couple of questions:
1. What should the `Readable.toWeb` and `Readable.fromWeb` return if the `webStreamAdapters` is an empty object?
2. What is the purpose of the `newStreamReadableFromReadableStream` and `newReadableStreamFromStreamReadable` functions, as I don't see them defined anywhere in the codebase?

mcollina:
Those are node core functions: https://nodejs.org/api/stream.html#streamreadablefromwebreadablestream-options

[...]

PR #506, opened 2023-02-26
fix: removes `Readable.fromWeb` and `Readable.toWeb` methods 

RishabhKodes:
fixes: https://github.com/nodejs/readable-stream/issues/482

The `Readable.fromWeb` and `Readable.toWeb` methods were no longer being used from here (instead being imported from `require(stream')` method).

This pr removes both methods during the build when the module is extracted from Node.js.

### DIFF
--- a/build/build.mjs
+++ b/build/build.mjs
@@ -1,5 +1,5 @@
 import { transform } from '@babel/core'
-import { createReadStream, writeFileSync } from 'node:fs'
+import { createReadStream } from 'node:fs'
 import { mkdir, readdir, readFile, rm, writeFile } from 'node:fs/promises'
 import { dirname, resolve } from 'node:path'
 import process from 'node:process'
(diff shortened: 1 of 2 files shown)


mcollina:
Can you move those to the replacements.mjs file?

RishabhKodes:
> Can you move those to the replacements.mjs file?

You mean the whole 

> Can you move those to the replacements.mjs file?

Do you mean moving the code changes I made, or the doing something like this https://github.com/nodejs/readable-stream/blob/main/build/build.mjs#L94 ?

mcollina:
Using the regular expressions instead.

RishabhKodes:
@mcollina I've moved the changes to `replacements.mjs` and put the code inside a function and then imported it in the `build.mjs` file, PTAL let me know if that's fine.

mcollina (changes requested):
What I was trying to say is to use the system that's already in place and not add new code, just replacements via regexps.

mcollina (approved):
lgtm

benjamingr (approved):

RishabhKodes:
@benjamingr @mcollina the `Node.js / Node.js (windows-latest, 18.x)` test is still failing after the re-run.

RishabhKodes:
@mcollina @benjamingr, an update on this pr? Has been idle for a long time.

mcollina:
Sorry about it, it was buried down my github notifications.

PR #508, opened 2023-03-01
Fix es5 compatibility

### LINKED ISSUE #507: 3.6.1 breaks ES5 compatibility
Our ES5 build breaks when `readable-stream` gets updated to 3.6.1.

This commit precisely introduces ES6 syntax, such as `const` keywords, class methods shorthands…

https://github.com/nodejs/readable-stream/commit/fec7fe9869a1f46bee0ce1603538059b780fe25a

I am aware that ES5 is a somewhat old thing, however, in my understanding, an EcmaScript version bump should not occur in a patch version.

mcollina:
Fixes #507

### DIFF
--- a/.babelrc
+++ b/.babelrc
@@ -2,9 +2,7 @@
   "presets": [
     ["@babel/preset-env", {
       "targets": [
-        "last 2 versions",
-        "not dead",
-        "node 6.0"
+        "node 0.10"
       ],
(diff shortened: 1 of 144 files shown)


benjamingr (approved):
RSLGTM

MonsieurLanza:
LGTM

PR #515, opened 2023-05-08
Update to Node v18.16.0

mcollina:


mcollina:
cc @ShogunPanda

ShogunPanda (approved):
LGTM!
Make sure you run `npm run format` (or similar, can't check the code now) as I see some file with unneeded changes.

Issue #516, opened 2023-05-10
When upgrading from `4.3.0` to `4.4.0`, using `esbuild --bundle` to build will report an error `Could not resolve "stream"`


leizongmin:
This is an example file `test.js`:

```js
require('readable-stream');
```

Run the following commands (esbuild@0.17.12):

```bash
esbuild --bundle test.js
```

It outputs:

```
✘ [ERROR] Could not resolve "stream"

    node_modules/.pnpm/readable-stream@4.4.0/node_modules/readable-stream/lib/stream/promises.js:7:8:
      7 │ require('stream')
        ╵         ~~~~~~~~

  The package "stream" wasn't found on the file system but is built into node. Are you trying to
  bundle for node? You can use "--platform=node" to do that, which will remove this error.

1 error
```

If I use version `4.3.0`, it will output correctly.

kanongil:
This seems like another instance of the issue in #513, though this exposes it as a regression!

leizongmin:
The problem was introduced from this commit https://github.com/nodejs/readable-stream/commit/b345071487c47b3c7b96007b26ea8bbe3ea97d15#diff-241fa22640048b6a147dec92f1ebf253043a1c4c342c7e52402112eb3c58f8fbR7

https://github.com/nodejs/readable-stream/blob/b345071487c47b3c7b96007b26ea8bbe3ea97d15/lib/stream/promises.js#L7

I have another question: why `require` this module, but not use any properties or methods of `stream`?

jeswr:
In order to avoid such regressions in the future can I suggest adding a test to this library which webpacks

import-all-test.js
```js
import * as all from 'readable-stream'
```

using the config

webpack.config.js
```js
const path = require("path");
module.exports = {
  mode: "development",
  entry: "./integration-tests/import-all-test.js",
  output: {
    path: path.resolve(__dirname, "integration-tests", "dist"),
    filename: "import-all-test.js",
  },
};
```

The test would just be running `npm run test:build` where that script is defined as `"test:build": "webpack"`

mcollina:
Would you like to send a PR to fix this (adjust the build scripts) or just with the test?

jeswr:
https://github.com/nodejs/readable-stream/pull/517 should cover the main point - I will see if I can make that config more restrictive after the issue is patched.

PR #517, opened 2023-05-13
chore: don't polyfill stream in webpack.browser.config.mjs

jeswr:
This PR does 2 things:
 - Cleans up the existing webpack browser config by removing alias' and plugins that are no longer required.
 - Tests for regression in #516 by adding a new webpack config to import what is effectively the file `import * as all from 'readable-stream'`

### DIFF
--- a/src/test/browser/fixtures/webpack.browser.config.mjs
+++ b/src/test/browser/fixtures/webpack.browser.config.mjs
@@ -20,15 +20,8 @@ export default {
       banner: 'function setImmediate(fn, ...args) { setTimeout(() => fn(...args), 1) }',
       raw: true
     }),
-    new webpack.ProvidePlugin({
-      process: require.resolve('process')
-    })
   ],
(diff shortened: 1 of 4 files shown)


jeswr:
Was Webpacking to fail with this change. Will investigate by Monday at the latest.

jeswr:
Should be ready to go now @mcollina (I don't have permissions to mark you as a reviewer so pinging you instead).

mcollina:
I thought it was a simple fix, but it's actually harder than the time I currently have available to fix it. @jeswr would you mind to identify a fix?

@ShogunPanda could you take a look as well?

jeswr:
Unfortunately I'm quite time poor myself at present @mcollina, so it is unlikely I will have a chance in the next 2 weeks to look into it.

If the issue hasn't been fixed after that - or if I find time beforehand - then I'll take a look.

PR #518, opened 2023-05-14
chore: add node 20 to CI checks

jeswr:
Noticed they were missing when looking into #517


jeswr:
I'm noticing that on the CI in my repo one of the tests are timing out on Node 16 on Windows for some reason (see https://github.com/jeswr/readable-stream/actions/runs/4973128200/jobs/8898847989).

I won't dig into it as it seems unrelated to this change, but wanted to note it nonetheless.

PR #519, opened 2023-05-15
fix: remove unused `require('stream')` and extend webpack tests

### LINKED ISSUE #516: When upgrading from `4.3.0` to `4.4.0`, using `esbuild --bundle` to build will report an error `Could not resolve "stream"`

jeswr:
Closes #516 

Potentially supercedes #517 by also including a fix; which is removing `require('stream')` in https://github.com/jeswr/readable-stream/blob/3caace80d34bacaeb3d2aa760f3bd0fbdd9ae3d3/lib/stream/promises.js.

However [this comment](https://github.com/nodejs/readable-stream/pull/517#issuecomment-1547427364) makes me suspicious of whether this one-liner is the desired fix - hence why I have opened this as a separate PR.

### DIFF
--- a/lib/stream/promises.js
+++ b/lib/stream/promises.js
@@ -4,7 +4,7 @@ const { ArrayPrototypePop, Promise } = require('../ours/primordials')
 const { isIterable, isNodeStream, isWebStream } = require('../internal/streams/utils')
 const { pipelineImpl: pl } = require('../internal/streams/pipeline')
 const { finished } = require('../internal/streams/end-of-stream')
-require('stream')
+
 function pipeline(...streams) {
   return new Promise((resolve, reject) => {
     let signal
--- a/src/test/browser/fixtures/webpack.browser.config.mjs
+++ b/src/test/browser/fixtures/webpack.browser.config.mjs
@@ -20,15 +20,8 @@ export default {
       banner: 'function setImmediate(fn, ...args) { setTimeout(() => fn(...args), 1) }',
       raw: true
     }),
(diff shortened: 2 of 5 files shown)


lib/stream/promises.js
```diff
@@ -4,7 +4,7 @@ const { ArrayPrototypePop, Promise } = require('../ours/primordials')
 const { isIterable, isNodeStream, isWebStream } = require('../internal/streams/utils')
 const { pipelineImpl: pl } = require('../internal/streams/pipeline')
 const { finished } = require('../internal/streams/end-of-stream')
-require('stream')
```
mcollina, line 7: This needs to come from the build step, not modified here.

jeswr: Ah - didn't realise that the contents of `lib/` were generated directly from the core nodejs codebase.

#520 should solve this

PR #520, opened 2023-05-15
fix: don't require `stream` package in codebase

### LINKED ISSUE #516: When upgrading from `4.3.0` to `4.4.0`, using `esbuild --bundle` to build will report an error `Could not resolve "stream"`

jeswr:
Closes #516.

Supercedes #517 and #519.

This replaces `require('stream')` with `require('../../lib/stream.js')` in the build script so that the `stream` module is never imported.

Since #517 I've moved the regression test to a new folder in the root since the other test fixtures are being copied from the nodejs codebase; wheras the one I add is specific to `readable-stream`.

### DIFF
--- a/readable-stream-test/runner-prepare.mjs
+++ b/readable-stream-test/runner-prepare.mjs
@@ -72,6 +72,6 @@ async function main() {
 }
 
 main().catch((e) => {
-  process.error(e)
+  error(e)
   process.exit(1)
 })
--- a/readable-stream-test/runner-prepare.mjs
+++ b/readable-stream-test/runner-prepare.mjs
@@ -1,0 +1,77 @@
+import { exec } from 'child_process'
+import util from '../lib/ours/util.js'
+
+function info(message) {
+  console.log(`\x1b[34m[INFO]\x1b[0m ${message}`)
+}
+
+function error(message) {
+  console.log(`\x1b[31m[INFO]\x1b[0m ${message}`)
+}
+
+async function run(command) {
+  info(`Executing \x1b[33m${command}\x1b[0m ...`)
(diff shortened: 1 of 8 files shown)


readable-stream-test/runner-prepare.mjs
```diff
@@ -0,0 +1,77 @@
+import { exec } from 'child_process'
```
mcollina, line 1: why is this needed?

jeswr: See line 27 (note this is a copy /paste of a subset of the code in the nodejs fixtures)

mcollina (approved):
lgtm

jeswr:
Bump @vweevers @benjamingr

jeswr:
@mcollina it doesn't look like the others have time to review so can this be released?

rubensworks:
Would it be possible to merge this package? Many of our downstream users are running into issues because of this

Happy to help where possible!

mcollina (approved):
lgtm

PR #471, opened 2022-04-15
Update to Node 18.0.0

ShogunPanda:
Summary:

1. Rewrite the build system reducing dependency.
2. Drop compatiblity for Node < 12.0 and Internet Explorer. Only active Node versions and ever-green browsers are supported.
3. Dropped most dependency (note: once we decide to drop support for Node 12 and Node 14, no dependency besides `object-inspect`, which is for browsers, will be needed)
4. Updated the exported `stream` API to Node 18.0.0.

PR #522, opened 2023-06-30
fix: add string decoder as a dependency

jeswr:
Adds `string_decoder` as a dependency as it is imported [here](https://github.com/nodejs/readable-stream/blob/ec65f43d95b8dd505f9059e0bd35aac309fdf19e/lib/internal/streams/readable.js#L66).

The `string_decoder` package happened to already be a devDependency of `readable-stream` and hence was not caught as missing by the recently added [webpacking tests](https://github.com/nodejs/readable-stream/tree/main/readable-stream-test) which have access to the devDependencies. Ideally a test should be added where they do not have access to these.

It appears to me that this used to be a dependency but that was removed in https://github.com/nodejs/readable-stream/commit/81c52641299e5ea3a5e37179967bb13691cad081#diff-7ae45ad102eab3b6d7e7896acd08c427a9b25b346470d7bc6507b6481575d519 - it is not clear to me why as a comment was made about this at the time https://github.com/nodejs/readable-stream/pull/471#discussion_r862719133.

### DIFF
--- a/.github/workflows/bundlers.yml
+++ b/.github/workflows/bundlers.yml
@@ -17,7 +17,9 @@ jobs:
         bundler: ['browserify', 'esbuild', 'rollup', 'webpack']
         exclude:
           - os: windows-latest
-            node-version: [12.x, 14.x]
+            node-version: 12.x
+          - os: windows-latest
+            node-version: 14.x
     steps:
       - name: Checkout
         uses: actions/checkout@v3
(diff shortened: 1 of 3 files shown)


mcollina:
Can you skip those failing jobs in our workflow file? Unfortunately GHA is broken on those old runtimes.

mcollina (commented):
There is an `exclude` attribute that you can add to the matrix instead.

mcollina (approved):
lgtm

PR #530, opened 2023-12-18
Use globalThis instead of global

### LINKED ISSUE #529: v4.5.0: global is not defined
I just upgraded to readable-stream@4.5.0 and I am now getting this error (using readable-stream in the browser):

`global is not defined`

I tracked these changes back to [this commit](https://github.com/nodejs/readable-stream/commit/4c8e8e528b46b9dba6cd4385b2f8195214dcc682#diff-9b8bb809feb796e3edeadfc9bc0cc24ced47a83b6be73d2fb9d95c868e54d75dR5), where there are many references to `global` that did not exist before. I'm wondering if they should be changed to `globalThis` instead, so it will not crash in browsers (`global` is only globally defined in node).

Here is one of the affected places for quick reference:
https://github.com/nodejs/readable-stream/blob/700d28b674a440ee939d0d88491e9040c45de642/lib/ours/util.js#L5-L6

mcollina:
Fixes #529

### DIFF
--- a/lib/ours/util.js
+++ b/lib/ours/util.js
@@ -2,8 +2,8 @@
 
 const bufferModule = require('buffer')
 const { kResistStopPropagation, SymbolDispose } = require('./primordials')
-const AbortSignal = global.AbortSignal || require('abort-controller').AbortSignal
-const AbortController = global.AbortController || require('abort-controller').AbortController
+const AbortSignal = globalThis.AbortSignal || require('abort-controller').AbortSignal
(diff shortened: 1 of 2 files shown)


mofux:
There's also another occurence in `src/util.js`:
https://github.com/nodejs/readable-stream/blob/700d28b674a440ee939d0d88491e9040c45de642/src/util.js#L5-L6

However, not sure if this one also needs modification.

PR #497, opened 2022-11-17
fix: use workaround loading process-shim for browserify

bergos:
replaced `require('process')` with `require('process/')` to fix #496

The build step generated quite a big diff. I compared a few samples with the code in the https://github.com/nodejs/node repository, and the generated code looks good to me. I think there was a problem earlier on.

@mcollina Hints on how to solve this are very welcome. Is there a post-process step? Should I create a separate PR just for the diff generated by the build and then fix the process-shim issue?

PR #543, opened 2024-10-14
Circumvent issues with ESM process polyfills

### LINKED ISSUE #539: Incompatibility with Rollup/Vite process polyfill
Similar to #450, though in this case a polyfill is present. It looks like the ``process`` polyfill is an ES module with a default export that isn't being properly unwrapped by ``readable-stream``.

I've set a breakpoint on the line that is causing ``TypeError: e.nextTick is not a function``:
image

I can confirm that a ``nextTick`` implementation is available at both the ``default`` and ``process`` fields. Here my ``vite.config.ts``:
```ts
import {defineConfig} from "vite";
import {nodePolyfills} from "vite-plugin-node-polyfills";

export default defineConfig({
    mode: 'development',
    build: {
        sourcemap: true
    },
    plugins: [
        nodePolyfills() // also tried: { globals: { process: true } }
    ]
});
```

| Package | Version |
| :-- | :-- |
| ``readable-stream`` | ``^4.5.2`` |
| ``vite`` | ``^5.4.2`` |
| ``vite-plugin-node-polyfills`` | ``^0.22.0`` |
| ``rollup`` | ``^4.21.0`` |
| ``esbuild`` | ``^0.23.1`` |

WasabiThumb:
Would resolve #539. A hacky solution that may preferrably be made obsolete by resolution of [davidmyersdev/vite-plugin-node-polyfills#106](https://github.com/davidmyersdev/vite-plugin-node-polyfills/issues/106).

The root issue is that some ``node:process`` polyfills export an ES module that is not unwrapped at runtime. This change would introduce a patch that conditionally unwraps the module in the event that checks pass indicating that it is wrapped. Those checks are:
- ``require("node:process")["__esModule"] === true``
  - Intent is to quickly eliminate false positives
- ``"nextTick" in require("node:process")["default"] || "nextTick" in require("node:process")["process"]``
  - Indicates that the ``default`` or ``process`` property of the import is indeed intended to polyfill ``node:process``. The choice of ``nextTick`` here is arbitrary, however the method would be required for proper function of this package.

This also inherits the curioisty introduced by #497 where ``node:process`` is imported via ``process/``.

### DIFF
--- a/lib/internal/patches/process.js
+++ b/lib/internal/patches/process.js
@@ -1,0 +1,27 @@
+/* wraps the internal process module, circumventing issues with some polyfills (see #539) */
+
+/** @type {import('node:process')} */
+const process = ((base, esmKey, keys, isValid) => {
+  // check if top-level es module, in which case it may have a default export
+  if (esmKey in base && base[esmKey] === true) {
+    let candidate
+    for (const key of keys) {
+      if (!(key in base)) {
+        continue
+      }
+      candidate = base[key]
+      // sanity check
+      if (isValid(candidate)) {
+        return candidate
+      }
+    }
+  }
+  return base
+})(
+  require('process/'),
+  '__esModule',
+  ['default', 'process'],
+  (candidate) => 'nextTick' in candidate
+)
+
+module.exports = process
--- a/lib/internal/streams/destroy.js
+++ b/lib/internal/streams/destroy.js
@@ -2,7 +2,7 @@
 
 /* replacement start */
 
-const process = require('process/')
+const process = require('../patches/process')
 
 /* replacement end */
 
--- a/lib/internal/streams/duplexify.js
+++ b/lib/internal/streams/duplexify.js
@@ -1,6 +1,6 @@
 /* replacement start */
 
-const process = require('process/')
+const process = require('../patches/process')
 
 /* replacement end */
 
--- a/lib/internal/streams/end-of-stream.js
+++ b/lib/internal/streams/end-of-stream.js
@@ -1,6 +1,6 @@
 /* replacement start */
 
-const process = require('process/')
+const process = require('../patches/process')
 
 /* replacement end */
 // Ported from https://github.com/mafintosh/end-of-stream with
--- a/lib/internal/streams/from.js
+++ b/lib/internal/streams/from.js
@@ -2,7 +2,7 @@
 
 /* replacement start */
 
-const process = require('process/')
+const process = require('../patches/process')
 
 /* replacement end */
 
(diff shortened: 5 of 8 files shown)


mcollina (changes requested):
Thanks for opening a PR! Can you please add a unit test?

WasabiThumb:
> Thanks for opening a PR! Can you please add a unit test?

Sorry for the delay! I've added the task ``test:vite``. It was very painful to replicate the correct setup, but it works like a charm in the end. It's not as graceful as the other browser tests, mostly because I couldn't get playwright working on Arch, but I believe this is a decent starting point. Specifically, the use of ``ViteDevServer.openBrowser()`` instead of playwright may not play well with CI. However, IPC is performed using an endpoint on the dev server rather than console stdout; therefore migration should be easy for those that can get playwright to run should that be desired. On the other hand, every case that that would catch should also be caught by the existing ``rollup`` tests.

WasabiThumb:
This is strange- despite using [the reprex](https://github.com/WasabiThumb/reprex-539) as a reference, the test does not break when commenting out the patching functionality of the shim introduced by this PR. It's possible that the issue only crops up when the Vite source is TypeScript, since that's the only obvious difference. I may have to consider updating the test to use TypeScript- or perhaps throw this all out and consider it a TypeScript issue.

gsantiago:
@WasabiThumb any update?

WasabiThumb:
> @WasabiThumb any update?

Moving this along would require writing a test that would actually fail without the fix, perhaps by converting the existing test's Vite source to TypeScript as that is a likely component of this issue and seems to be present whenever it's reported. Since I'm not actively developing the project in which I ran into this issue right now, it might be a while until I am able to do that myself. If you want to get ahead of it, you can [fork my repo](https://github.com/WasabiThumb/readable-stream/tree/bug/vite-esm) and PR a test/modify the existing test such that it demonstrates this behavior; then I'll happily resolve the conflicts and request a review.

mesqueeb:
@mcollina isn't there any way you can alleviate the blockage of this PR. readable-stream doesn't work with vite based esm builds...

mcollina:
I have no time to volunteer to work on a comprehensive test suite for this bug. Without it, we'll keep releasing versions after versions to chase all the different tool combinations. If you are writing on behalf of a company and would like to sponsor this work, ping me via email.

mesqueeb:
@mcollina can't you just release it as a beta version? no need for the test suite, I can test the beta version in my app directly.

MattiasBuelens:
@mesqueeb You can always fork the package, or use something like [patch-package](https://www.npmjs.com/package/patch-package) to apply patches directly to the `readable-stream` code inside your `node_modules`.

Or even better, since you apparently ran into this issue yourself: try to make a minimal reproduction case that we can use for the test! 😉 As you can read in the thread, we're struggling to make a good test for this, so any help with that would be appreciated.

PR #544, opened 2024-11-30
fix: `internalStreamsWeakHandler` replacement rule

natebuckareff:
The replacement part of the rule incorrectly contains the escape `\\`, probably copy-pasted from the pattern part. Fix is to just remove it.


mcollina (approved):
lgtm

PR #548, opened 2024-12-31
Fix `string_decoder` import

### LINKED ISSUE #547: Library resolves to Node.js '`string_decoder`' module in Vercel Edge Environment
When trying to use the `Readable` exported by this library in a Vercel Edge Function one gets the following error:

```
Error during transformationError: The edge runtime does not support Node.js 'string_decoder' module.
```

jeswr:
Resolves #547

Have tested this fix in a vercel edge environment


lib/internal/streams/readable.js
```diff
@@ -65,7 +65,7 @@ const {
 } = require('../../ours/errors')
 const { validateObject } = require('../validators')
 const kPaused = Symbol('kPaused')
-const { StringDecoder } = require('string_decoder')
+const { StringDecoder } = require('string_decoder/')
```
MattiasBuelens, line 68: This file is copied directly from the Node sources using [build.mjs](https://github.com/nodejs/readable-stream/blob/main/build/build.mjs). As such, you shouldn't edit this file directly.

Instead, add a new replacement to the list in [replacements.mjs](https://github.com/nodejs/readable-stream/blob/fdfbf827bc29ebdcb26e93d3566891dc18f06d9e/build/replacements.mjs#L295), and then run `build.mjs` again.

MattiasBuelens (changes requested):

lib/internal/streams/readable.js
jeswr: Got it - what node version should I be building with?

jeswr: Ok I've rebuilt, and also put the node version as a config option in the package.json so it is clear what node version things were built from.

package.json
```diff
@@ -82,5 +82,10 @@
   },
   "engines": {
     "node": "^12.22.0 || ^14.17.0 || >=16.0.0"
+  },
+  "config": {
+    "node": {
+      "version": "18.20.5"
```
MattiasBuelens, line 88: Maybe we should just pass the Node version as an explicit parameter in [the `build` npm script]()?

```diff
  "scripts": {
-   "build": "node build/build.mjs",
+   "build": "node build/build.mjs 18.20.5",
```

jeswr: Done :)

README.md
```diff
@@ -11,9 +11,9 @@
 npm install readable-stream
 ```
 
-This package is a mirror of the streams implementations in Node.js 18.19.0.
```
MattiasBuelens, line 14: Okay, so turns out we were on Node 18.19.0 before, not 18.20.5. 😅 

We should definitely consider updating, but maybe keep that for a separate PR. 😉

jeswr: Done - there appears to still be a larger diff than I would expect; seemingly due to eslint using a different version as a result of a lack of lockfile

mcollina:
I see a lot of formatting changes. Why are those in?

MattiasBuelens:
@mcollina That's just the result of `npm run build && npm run format`. It looks like we haven't done that in a while... 🤷‍♂️

We should really have CI check that for us. I'll make a PR afterwards to run both Prettier and ESLint on CI.

mcollina (approved):
lgtm

MattiasBuelens (approved):

jeswr:
Thanks for reviewing @mcollina @MattiasBuelens . I do not have write access to this repository so one of you will need to perform the merge.

PR #542, opened 2024-10-08
fix: avoid using bigint literal syntax

mman:
Attempts to address https://github.com/nodejs/readable-stream/issues/541. After the discussion I have made a custom ESLint plugin that exposes `no-big-int` rule to catch usage of the bigint `n` literal as part of `npm run lint`. In order to actually make the local ESLint plugin / local ESLint rule work, I have to also migrate the ESLint configuration to the new flat file format.

Running the `npm run lint` against `main` reports this (note that there are unrelated linting errors that should probably be addressed in a separate PR).

```
€ npm run lint                              

> readable-stream@4.5.2 lint
> eslint src


/Users/mman/Developer/victronenergy/readable-stream/src/errors.js
  339:19  error  ES2020 `bigint` syntax is forbidden  local/no-big-int
  339:25  error  ES2020 `bigint` syntax is forbidden  local/no-big-int
  339:42  error  ES2020 `bigint` syntax is forbidden  local/no-big-int
  339:48  error  ES2020 `bigint` syntax is forbidden  local/no-big-int

/Users/mman/Developer/victronenergy/readable-stream/src/primordials.js
  106:3  warning  Expected property shorthand  object-shorthand

/Users/mman/Developer/victronenergy/readable-stream/src/util.js
   27:15   error  'ERR_INVALID_ARG_TYPE' is not defined  no-undef
   27:64   error  Extra semicolon                        semi
   29:2    error  Extra semicolon                        semi
   32:5    error  Expected { after 'if' condition        curly
   32:15   error  'ERR_INVALID_ARG_TYPE' is not defined  no-undef
   32:60   error  Extra semicolon                        semi
   33:2    error  Extra semicolon                        semi
  158:17   error  'ERR_INVALID_ARG_TYPE' is not defined  no-undef
  158:70   error  Extra semicolon                        semi
  160:42   error  Extra semicolon                        semi
  161:43   error  Extra semicolon                        semi
  163:28   error  Extra semicolon                        semi
  165:39   error  Extra semicolon                        semi
  167:114  error  Extra semicolon                        semi
  169:54   error  Extra semicolon                        semi
  170:8    error  Extra semicolon                        semi
  175:32   error  Extra semicolon                        semi
  176:8    error  Unexpected trailing comma              comma-dangle
  177:6    error  Extra semicolon                        semi
```

Running it against my fork looks clean:

```
€ npm run lint

> readable-stream@4.5.2 lint
> eslint src


[...]

PR #549, opened 2024-12-31
Update build for #542

### LINKED ISSUE #541: FR: Avoid using bigint literal syntax to target older browsers
One of my projects targets very old specialised browsers, and when `readable-stream` dependency is brought in the webpack produced code fails to run because of `SyntaxError` here:

https://github.com/nodejs/readable-stream/blob/a2e9aedf4aeee4a5e4d8efcb175edb67e2817eaa/src/errors.js#L339-L340

Turns out that specialised browser based on WebKit can not parse bigint literals.

I'm not sure if I am missing something but I was not able to find a solution to transpile the bigint literals correctly to target older browsers.

But turns out that the fix could be to avoid bigint literal and use `BigInt` constructor.

MattiasBuelens:
I noticed that the changes from #542 only affected `src/`, but not `lib/`. As such, [the latest 4.6.0 release](https://github.com/nodejs/readable-stream/releases/tag/v4.6.0) doesn't actually fix #541, since it still has [the bigint literal fyntax in `lib/ours/errors.js`](https://unpkg.com/browse/readable-stream@4.6.0/lib/ours/errors.js).

I wanted to fix this by rerunning `node build/build.mjs` locally. However, that surfaced a couple of more issues, which this PR aims to fix as well:
1. The regex for inserting headers was looking for `"use strict"` *only* on the very first line. However, most source files start with a copyright comment, so [the `"use strict"` directive only appears a few lines further](https://github.com/nodejs/node/blob/v18.19.0/lib/internal/streams/readable.js#L22). I fixed it by turning on the multiline flag on the regex.
2. `src/util.js` was using `ERR_INVALID_ARG_TYPE` without `require`ing it from `src/errors.js`. I added the missing require.
3. I reran Prettier and ESLint.
4. Finally, I reran `node build/build.mjs 18.19.0`.

I think we should have a test on our CI to verify that `lib/` has been correctly built using our build script. It's not always obvious to new contributors that `lib/` is generated code [(see review comment)](https://github.com/nodejs/readable-stream/pull/548#discussion_r1900059594), and it should (hopefully) prevent mistakes like #542 in the future.

### DIFF
--- a/lib/ours/util.js
+++ b/lib/ours/util.js
@@ -5,7 +5,7 @@ const { format, inspect } = require('./util/inspect')
 const {
   codes: { ERR_INVALID_ARG_TYPE }
 } = require('./errors')
-const { kResistStopPropagation, SymbolDispose } = require('./primordials')
+const { kResistStopPropagation, AggregateError, SymbolDispose } = require('./primordials')
 const AbortSignal = globalThis.AbortSignal || require('abort-controller').AbortSignal
 const AbortController = globalThis.AbortController || require('abort-controller').AbortController
 const AsyncFunction = Object.getPrototypeOf(async function () {}).constructor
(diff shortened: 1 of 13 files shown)


lib/ours/util.js
```diff
@@ -1,6 +1,9 @@
 'use strict'
 
 const bufferModule = require('buffer')
+const {
+  codes: { ERR_INVALID_ARG_TYPE }
+} = require('./errors')
```
MattiasBuelens, line 6: This creates a cyclic dependency between `ours/util.js` and `ours/errors.js`... 😖

I'll try to split up these files to remove cycle.

mcollina (approved):
lgtm

mcollina:
Any improvement to CI is very welcomed!

PR #550, opened 2025-01-01
Run lint on CI

MattiasBuelens:
This adds a new "Lint" workflow which runs on all branches and pull requests and performs some checks:
* Runs `npm run build`, and checks that the generated files exactly match those in the `lib/` directory.
* Runs ESLint, to check that there are no lint errors.
* Runs Prettier, to check that there are no code formatting issues.


mcollina (approved):
lgtm

PR #554, opened 2025-08-11
Fix broken repository URL in package.json

ExplodingCabbage:
GitHub hasn't supported the git protocol [since 2022](https://github.blog/security/application-security/improving-git-protocol-security-github/#no-more-unauthenticated-git), so this URL is broken.

(This is a common error affecting many popular npm packages; I'm opening PRs on several projects to fix it. I encourage anyone reading this to check any packages they maintain and implement the same fix if appropriate!)


lbesecker195 (approved):
GitHub removed unauthenticated `git://` support in 2022, so the old URL cannot be cloned. HTTPS is what npm's docs recommend for the `repository` field. LGTM.

Issue #106, opened 2015-01-27
io.js readable-streams WG meeting


chrisdickinson:
~~I've put together a [doodle](https://doodle.com/5srxwvsk6sbz2evz) that proposes an hour long meeting for wednesday (1/28), thursday (1/29), or friday (1/30). If those dates don't work, I can reopen for next week.~~

**UTC Wed 30-Jan-15 17:00:**

San Francisco: Wed 28-Jan-15 09:00
Amsterdam: Wed 28-Jan-15 18:00
Moscow: Wed 28-Jan-15 20:00
Sydney: Thu 29-Jan-15 04:00

cc @iojs/streams 

**Public Youtube feed**: http://www.youtube.com/watch?v=I9nDOSGfwZg
**Agenda and minutes doc**, (eventually) editable by WG: https://docs.google.com/document/d/17aTgLnjMXIrfjgNaTUnHQO7m3xgzHR2VXBTmi03Qii4/edit#

**Agenda So Far**:

Extracted from https://github.com/iojs/readable-stream/labels/wg-agenda prior to meeting.

* adopt a charter [#105](https://github.com/iojs/readable-stream/issues/105)
* release and versioning strategy [#101](https://github.com/iojs/readable-stream/issues/101)
* simpler stream creation [#102](https://github.com/iojs/readable-stream/issues/102)
* proposal: deprecate implicit flowing of streams [#99](https://github.com/iojs/readable-stream/issues/99)

mikeal:
could I sit in on this one as well?

sonewman:
I have put some times in, the difficulty I have is that the best times for you guys is mostly 7pm onwards for me which is a little tricky since I have to commute from London, and when i get back I have to look after my daughter. I'm not sure what the google hangouts app is like for iPhone, but I guess i'll let you know :smile:

rvagg:
![screen shot 2015-01-28 at 9 08 33 am](https://cloud.githubusercontent.com/assets/495647/5928199/5716e476-a6cd-11e4-82f1-7c2089860270.png)

You're killing me @sonewman, your availability is all 5am for me.

Actually, hangouts is great on mobile, and you don't have to have video turned on of course.

sonewman:
@rvagg ahh, we have a 11 hour time difference :stuck_out_tongue_winking_eye: I could do it in work hours, since my work is pretty flexible about this sort of thing, but I can't select anything between 4am and 6pm. 

[...]

PR #557, opened 2026-07-03
fix: unwrap ESM-style process polyfills in internal streams (fixes #539)

### LINKED ISSUE #539: Incompatibility with Rollup/Vite process polyfill

smessie:
## Summary

Some browser bundlers/polyfill setups expose the `process/` polyfill as an ES module with a `default` export (i.e. `{ __esModule: true, default: <process> }`). The internal streams code does `const process = require('process/')` and then uses `process.nextTick(...)`, which throws in that scenario because the module namespace object has no `nextTick`.

This routes the internal `process/` imports through a small shim that unwraps such ESM-style polyfills (returning the first candidate exposing `nextTick`) while leaving normal CommonJS polyfills untouched.

Fixes #539.
This PR follows a similar approach as #543.

## Changes

- Add `lib/internal/shims/process.js` (generated from `src/`) that unwraps an ESM-default `process/` polyfill, validating candidates via `'nextTick' in candidate`, and falls back to the module as-is otherwise.
- Route the internal streams (`destroy`, `duplexify`, `end-of-stream`, `from`, `pipeline`, `readable`, `writable`) through the shim instead of `require('process/')` directly.
- Wired through the repo's build/replacement generation so both the source and generated outputs are consistent.

## Tests

- Adds `test/ours/test-process-shim.js` verifying the shim resolves the real process object (with `nextTick`) when given an ESM-default-shaped polyfill, and returns a plain process-like object unchanged.
- `npm test`, `npm run lint`, `npm run test:format`, and the browserify bundle test all pass.

## Context

This fix is currently carried as a `patch-package` patch in [Comunica](https://github.com/comunica/comunica) to make it work in the browser; upstreaming it here removes the need for the patch. See https://github.com/comunica/comunica/pull/1724.


smessie:
Any chance we can have a look at this @mcollina ?

mcollina (approved):
lgtm

mcollina:
ci looks red

smessie:
@mcollina This is because the CI still uses unsupported Node.js versions.
```
Run ./node_modules/.bin/playwright install firefox
You are running Node.js 16.20.2.
Playwright requires Node.js 18 or higher. 
Please update your version of Node.js.
Error: Process completed with exit code 1.
```
I've pushed a commit that updates the CI to the [latest LTS versions](https://nodejs.org/en/about/previous-releases).

smessie:
Does this look good to you @mcollina or do you prefer another way to solve this CI problem?

smessie:
Any chance we can get this merged @mcollina ?

.github/workflows/node.yml
```diff
@@ -13,16 +13,7 @@ jobs:
...
-        exclude:
-          - os: windows-latest
-            node-version: 12.x
-          - os: windows-latest
-            node-version: 14.x
-          - os: macos-latest
-            node-version: 12.x
-          - os: macos-latest
-            node-version: 14.x
+        node-version: [22.x, 24.x, 26.x]
```
mcollina: Unfortunately, we still needs to test 12, 14, 16, 18, 20 and 21.

smessie:
The failing CI is not related to this PR, yet I pushed a commit that solves it. I dropped my previous commit such that it is testing against the old Node.js versions again.
The new commit updates the browser tests to Node.js 22 as playwright, used for these tests, no longer supports Node.js 16, so any other PR would face the same issue.

Does this suffice, @mcollina? If not, please be specific about the path you want to take. I’ll then determine if I still have the time to investigate this, or if I’ll have to abandon this effort.

smessie:
Any word on this @mcollina ?

smessie:
@mcollina I kept the old Node.js versions in the test matrices, as requested. Two extra commits were needed to get CI green.

**1. `ci: run browser tests on Node.js 22` (`browsers.yml`)**

`playwright` is declared as `^1.21.1`, so a fresh `npm install` in CI now resolves to 1.62.0, which requires Node.js 20 or higher. The `Install Browser` step aborted on Node.js 16 before any test ran. Only `browsers.yml` is bumped. `node.yml` and `bundlers.yml` keep testing Node.js 12 through 21.

**2. `fix: declare ajv explicitly so webpack bundling works on npm 6/7` (`package.json`)**

`webpack` pulls in `schema-utils@4`, which depends on `ajv-keywords@5` and its peer dependency `ajv@^8.8.2`. npm 6 and 7, shipped with Node.js 12 and 14, do not install peer dependencies, so `ajv-keywords` ends up hoisted next to the `ajv@6` that eslint requires, and fails with:

```
[webpack-cli] Error: Cannot find module 'ajv/dist/compile/codegen'
```

That breaks `npm run test:prepare webpack` on those two versions. Declaring `ajv@^8.8.2` hoists v8 to the root of the tree and nests eslint's v6 underneath it. npm 8 and later install the peer dependency already, so nothing changes there.

This one is not caused by this PR. The same failure reproduces on `main` with Node.js 14 and npm 6, so the `ubuntu-latest, 12.x, webpack` and `ubuntu-latest, 14.x, webpack` legs of the bundlers workflow were already red.

Since workflow runs here need maintainer approval, I verified everything on my fork:

* Bundlers before the ajv commit, 42/44, both webpack legs on Node.js 12 and 14 failing: https://github.com/smessie/readable-stream/actions/runs/31486989796
* Bundlers after, 44/44: https://github.com/smessie/readable-stream/actions/runs/31492474993
* Browsers on Node.js 22, 40/40: https://github.com/smessie/readable-stream/actions/runs/31492474977

Full matrix on the final commit is green: Browsers 40/40, Bundlers 44/44, Node.js 14/14, Lint 1/1.

Happy to split the ajv change into a separate PR if you prefer to keep this one focused on the process polyfill fix.

smessie:
Any chance we can get this merged @mcollina ?

WasabiThumb:
This is essentially identical to #543 but with a less ambitious test case that doesn't attempt to verify whether the shim is necessary in the wild. The reason #543 got blocked is because I gave up on creating a test that fails when the shim is not enabled. I think that is going to be vital in this case, going off [Collina's word](https://github.com/nodejs/readable-stream/issues/539#issuecomment-2411638348). Until then individuals will continue to find these issues/PRs and substitute one of our forks into their ``package.json``, borrowing the fix in its hacky state and knowing the risks.

lib/internal/shims/process.js
```diff
@@ -0,0 +1,19 @@
...
+      }
+      candidate = base[key]
+      if (isValid(candidate)) {
+        return candidate
+      }
+    }
+  }
+  return base
+})(require('process/'), '__esModule', ['default', 'process'], (candidate) => 'nextTick' in candidate)
+module.exports = process
```
WasabiThumb, line 19: Is this version of the code not less readable than [the original](https://github.com/nodejs/readable-stream/pull/543/changes#diff-ec1a6cfd4a24b5ac665c7fd72259a51788e237b781515ca24a50bedf805dc61cR1)? Why choose to strip comments and increase line length in this way?

src/internal/shims/process.js
```diff
@@ -0,0 +1,21 @@
+'use strict'
```
WasabiThumb, line 1: Duplicate code is duplicate

smessie: That's the repo's layout, not duplication. `src/` is authored, `lib/` is generated: `build/build.mjs` runs `rm -rf lib` and regenerates from `src/`, which is why `src/util.js`/`lib/ours/util.js`, `src/errors.js`/`lib/ours/errors.js` etc. all exist in both trees. The shim needs the same treatment, plus the `contents.push(['lib/internal/shims/process.js', ...])` line in `build/build.mjs` and the `build/headers.mjs` change, otherwise the next `npm run build` deletes the shim and regenerates stream files requiring a module that no longer exists.

lib/internal/shims/process.js
smessie: That file is generated output — `build/build.mjs` reads `src/internal/shims/process.js` and runs it through babel + prettier, so its formatting isn't hand-written.

On line length specifically: the repo's prettier config is `printWidth: 120`, and running it over the formatting in #543 collapses the argument list to exactly this line. `npm run test:format` (`prettier -c src lib test`) fails on #543's version as committed. So this is what repo CI produces either way.

smessie:
The bar mcollina set on #539 was "I'll be happy to land a PR covered by tests"; the vite/playwright line came the day after, answering your own question about what a test would look like. This PR adds four tests covering the shim's behaviour and is approved.

An end-to-end test is a bigger piece of work than this fix: `main` has fixtures for esbuild, rollup and webpack only, so it means a new vite + playwright harness. It would also pin CI to the behaviour of `vite-plugin-node-polyfills`, whose own issue (#106) is still open — when that's fixed upstream, such a test stops demonstrating anything.

Happy to see that harness land as its own PR. I don't think it should block a fix that people are currently patching in by hand.

